PR #4, opened 2017-05-11
Added Behavioural Tests for Selected End Points - Message Put

MIslam92:


tests/behavioural/features/steps/message_put.py
```diff
@@ -339,11 +341,10 @@ def step_impl(context):
 
 
 @then("the status of that message changes to 'read'")
-def step_impl(context,step):
+def step_impl(context):
     response_get = app.test_client().get("http://localhost:5050/message/{0}".format(context.msg_id), headers=headers)
     data_get = json.loads(response_get.data)
     nose.tools.assert_true("ARCHIVE" not in data_get['labels'])
-    return step.error_message
```
AndrewTorrance: `not clear what archive is doing here ?

tests/behavioural/features/steps/message_put.py
```diff
@@ -373,14 +375,31 @@ def step_impl(context):
 def step_impl(context):
     headers['user_urn'] = 'respondent.123'
     data['urn_from'] = headers['user_urn']
-    response = app.test_client().post("http://localhost:5050/draft/save", data=json.dumps(data), headers=headers)
+    response = app.test_client().post("http://localhost:5050/draft/save",
+                                      data=json.dumps(data), headers=headers)
     post_resp = json.loads(response.data)
     context.msg_id = post_resp['msg_id']
```
AndrewTorrance: This only checks the message id , not if any of the data has changed as expected ?

tests/behavioural/features/message_get.feature
```diff
@@ -41,3 +41,9 @@ Feature: Message get by ID Endpoint
     Given an external user has created and saved a draft message
     When the external user navigates to draft messages and opens the draft message
     Then the draft message should be displayed
+
+  @ignore
+  Scenario: As a user when I send a invalid message with bad syntax
+    Given a user is sending a message
```
AndrewTorrance: The scenario should describe the test , there should be a piece on the end of the Scenario saying what happens

AndrewTorrance: Also we should have a test for each specific bad parameter . SO it should be one test per field in error . There is a good example of doing this with parameters in messages_get on master

tests/behavioural/features/steps/message_put.py
```diff
@@ -339,11 +341,10 @@ def step_impl(context):
 
 
 @then("the status of that message changes to 'read'")
-def step_impl(context,step):
+def step_impl(context):
     response_get = app.test_client().get("http://localhost:5050/message/{0}".format(context.msg_id), headers=headers)
     data_get = json.loads(response_get.data)
     nose.tools.assert_true("ARCHIVE" not in data_get['labels'])
-    return step.error_message
```
AndrewTorrance: Not sure why archive here ?

PR #7, opened 2017-05-15
Ts 174 restructure unit tests

MIslam92:


tests/app/test_authenticator.py
```diff
@@ -64,3 +64,31 @@ def test_authentication_jwt_not_encrypted_fail(self):
 
         with self.assertRaises(BadRequest):
             check_jwt(encode(data))
+
+    def test_authenticate_pass(self):
+        """Authenticate request using authenticate function and with correct header data"""
```
AndrewTorrance: test names should be more explicit not just pass or fail

PR #8, opened 2017-05-18
Ts 167 draft edit checks

tej99:


app/resources/drafts.py
```diff
@@ -82,6 +82,13 @@ def put(self, draft_id):
         is_draft = self.check_valid_draft(draft_id)
         if is_draft is False:
             raise (BadRequest(description="Draft put requires valid draft"))
+
+        not_modified = self.etag_check(request.headers, is_draft[1])
+        if not_modified == False:
+            res = Response(response="Draft has been modified since last check", status=409,
+                           mimetype="text/html")
```
AndrewTorrance: check ? Surely it should be 'Draft has been modified since it was last read' not check

app/resources/drafts.py
```diff
@@ -101,22 +102,24 @@ def put(self, draft_id):
             return resp
 
     @staticmethod
-    def check_valid_draft(draft_id):
+    def check_valid_draft(draft_id, user_urn):
         """Check msg_id is that of a valid draft"""
```
AndrewTorrance: This is no longer just checking , its also getting so the function name and the code string are not correct . The name should reflect the function .  Maybe a get_draft and a separate is_valid_draft passing it the result of check draft ? . Or even a combined function with a better name ?

tests/behavioural/features/message_post.feature
```diff
@@ -55,6 +55,7 @@ Feature: Message Send Endpoint
     When the message is sent
     Then a 400 error status is returned
 
-
-
-
+  Scenario: When a message with the label of "Draft" is sent and another user is trying to send the same message return a 409
```
AndrewTorrance: We also need a test so that the etag is NOT checked in this same scenario if the POST is done without an etag

tests/behavioural/features/steps/message_post.py
```diff
@@ -141,6 +142,55 @@ def step_impl(context):
...
+@given('a draft message is posted')
+def step_impl(context):
+    if 'msg_id' in data:
+        del data['msg_id']
+
+    context.post_draft = app.test_client().post("http://localhost:5050/draft/save", data=json.dumps(data),
+                                                headers=headers)
+    #get etag from response using context
+    context.etag =  json.loads(context.post_draft.data)
+    headers['Etag'] = context.etag
```
AndrewTorrance: I dont think this is a valid Etag ? Shouldnt there be a function call / something that hashes the data ?

PR #12, opened 2017-05-30
D 133/get draft and drafts not returning urn to

tej99:


tests/app/test_app.py
```diff
@@ -132,8 +132,9 @@ def test_missing_thread_id_does_not_cause_exception(self):
         """posts to message send end point without 'thread_id'"""
         url = "http://localhost:5050/message/send"
 
-        test_message = {'urn_to': 'richard',
-                        'urn_from': 'torrance',
+        now = datetime.now(timezone.utc)
```
gemmai95: This lines seems to fail one of the checks. If you have a look at that it should be fine.

PR #50, opened 2017-07-06
Changed setting file to point to RAS Gov.uk notify and update template

emilioward:

Trello Card - [https://trello.com/c/4DKHDMGD](https://trello.com/c/4DKHDMGD)

Updated the settings.py file to point to the RAS gov.uk notify account and updated the template according to the trello card. 

**How to review**

Send a message and check that the email gets sent to my inbox (Not sure if we have to change that to a RAS email account but for testing left as is)


AndrewTorrance:
Has this been demo'd to Louise ?

emilioward:
Demo'd to louise and she seems very happy with the email

emilioward:
Updated confluence page with details - https://digitaleq.atlassian.net/wiki/display/RASB/Developer+set+up+for+GOV+Notify+Email+Notifications

You can easily add as many templates to a service and have multiple templates for any combination of emails/ SMS depending on your needs. All you need to do is grab the API details from the newly created template from Gov.UK Notify and plug those into the appropriate ENV within settings.py file (Service ID, API ID and modify the Template ID).

PR #67, opened 2017-07-31
Initial push new info end point - values hardcoded

AndrewTorrance:

https://trello.com/c/3irefzH7

Added new get info end point that currently returns constant JSON. The next stage will be to read this information externally 

**How to review**

### DIFF
--- a/app/application.py
+++ b/app/application.py
@@ -7,6 +7,7 @@ from app import settings
 from app.exception.exceptions import MessageSaveException
 from app.repository import database
 from app.resources.health import Health, DatabaseHealth, HealthDetails
+from app.resources.info import Info
 from app.resources.messages import MessageList, MessageSend, MessageById, MessageModifyById
 from app.authentication.authenticator import authenticate
 from app.resources.drafts import DraftSave, DraftById, DraftModifyById, DraftList
@@ -40,6 +41,7 @@ with app.app_context():
 api.add_resource(Health, '/health')
 api.add_resource(DatabaseHealth, '/health/db')
 api.add_resource(HealthDetails, '/health/details')
+api.add_resource(Info, '/info')
 api.add_resource(MessageList, '/messages')
 api.add_resource(MessageSend, '/message/send')
 api.add_resource(MessageById, '/message/<message_id>')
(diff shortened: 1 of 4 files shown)


app/application.py
```diff
@@ -54,14 +56,21 @@ def drop_database():
...
-    if request.endpoint is not None and 'health' not in request.endpoint and request.method != 'OPTIONS':
+    if _request_requires_authentication(request):
         log_request()
         res = authenticate(request.headers)
         if res != {'status': "ok"}:
             logger.debug("Failed to authenticate user")
             return res
 
 
+def _request_requires_authentication(req):
```
dcdarrell9: What is this 'req'? Wouldn't it be better to have the full name.

app/application.py
```diff
@@ -54,14 +56,21 @@ def drop_database():
 
 @app.before_request
 def before_request():
-    if request.endpoint is not None and 'health' not in request.endpoint and request.method != 'OPTIONS':
+    if _request_requires_authentication(request):
         log_request()
         res = authenticate(request.headers)
         if res != {'status': "ok"}:
             logger.debug("Failed to authenticate user")
             return res
```
dcdarrell9: Same with the 'res' as 'req'.

app/resources/info.py
```diff
@@ -0,0 +1,21 @@
...
+    """Rest endpoint to provide application information"""
+
+    @staticmethod
+    def get():
+        details = {'name': 'secure_message',
+                   'version': '0.0.1',
+                   'commit': 'not specified',
+                   'branch': 'not specified',
+                   'built': '01-01-1900 00:00:00.000'}
+
```
dcdarrell9: Does 'origin' need to be included?

app/application.py
AndrewTorrance: No because request is a global. Any use of the name request will mask the global . So the temptation is to not pass it and rely on the global . However if you do that then the function becomes difficult and messy to test . Perhaps a different name

app/resources/info.py
AndrewTorrance: Thanks , I missed that

PR #71, opened 2017-08-08
Feature/message validation

rjweeks:
Changes to the domain so that subject and body fields are not deleted or changed when they have an error.

### DIFF
--- a/app/validation/domain.py
+++ b/app/validation/domain.py
@@ -185,12 +185,12 @@ class DraftSchema(Schema):
     @validates("body")
     def validate_body(self, body):
         if body is not None:
-            self.validate_field_length(body, len(body), constants.MAX_BODY_LEN)
+            self.validate_field_length('body', len(body), constants.MAX_BODY_LEN, body)
 
     @validates("subject")
     def validate_subject(self, subject):
         if subject is not None:
-            self.validate_field_length(subject, len(subject), constants.MAX_SUBJECT_LEN)
+            self.validate_field_length('subject', len(subject), constants.MAX_SUBJECT_LEN, subject)
 
     @validates("thread_id")
     def validate_thread_id(self, thread_id):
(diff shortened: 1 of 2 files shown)


app/validation/domain.py
```diff
@@ -65,14 +65,14 @@ def validate_to_from_not_equal(self, data):
             raise ValidationError("msg_to and msg_from fields can not be the same.")
 
     # TODO: Validate UUID with mock party
-    @validates('msg_to')
+    @validates("msg_to")
     def validate_to(self, msg_to):
         for item in msg_to:
             self.validate_non_zero_field_length("msg_to", len(item), constants.MAX_TO_LEN)
             if msg_to != 'BRES' and not User.is_valid_user(item):
                 raise ValidationError("{0} is not a valid user.".format(item))
```
rhiGriff: Should this be consistent with the other ValidationError lines in regards to single quotes rather than double quotes?

app/validation/domain.py
```diff
@@ -65,14 +65,14 @@ def validate_to_from_not_equal(self, data):
             raise ValidationError("msg_to and msg_from fields can not be the same.")
```
rhiGriff: Should this be consistent with the other ValidationError lines in regards to single quotes rather than double quotes?

dcdarrell9:
I don't know why, but this is the error that I'm getting.

screen shot 2017-08-08 at 10 57 02

app/validation/domain.py
```diff
@@ -128,11 +128,10 @@ def validate_non_zero_field_length(self, field_name, length, max_field_len):
...
     @staticmethod
-    def validate_field_length(field_name, length, max_field_len):
+    def validate_field_length(field_name, length, max_field_len, data=None):
         if length > max_field_len:
             logger.debug("{0} field is too large {1}  max size: {2}".format(field_name, length, max_field_len))
-            raise ValidationError('{0} field length must not be greater than {1}.'.format(field_name, max_field_len))
-
+            raise ValidationError('{0} field length must not be greater than {1}.'.format(field_name, max_field_len), field_name, [], data)
 
 class DraftSchema(Schema):
```
andrewmil: Since DraftSchema is almost the same as the MessageSchema could we just make it a subclass of MessageSchema and overwrite the different functions?

app/validation/domain.py
rjweeks: So these decorators are consistent using "

app/validation/domain.py
rjweeks: Yes that is worth doing.

PR #73, opened 2017-08-11
Ts 232 use party service in addition to mocked version

AndrewTorrance:
Up to now the party service has been mocked . This change keeps that mock for test purposes but also integrates use of the actual party service . 

https://trello.com/c/HPEWccWq

- New services folder intended to hold code for accessing external services
- Created a new Service/mock toggle to enable code to switch .
- created  party object that application code should use
-Modified the ubiquitous use of 'BRES' to 2 new constants , one for BRES_USER and one for BRES_SURVEY 
-Changed name of data returned from party_service mock so that it aligns with the real party service (email=>emailAddress , surname=>lastName etc) 

- Added tests for the new mockservice toggles
-added new integration tests that need to be manually run for testing responses from the real party service

- Added authorisation check for get message by id so that a user cannot see a message if they are a respondent and not on the to or from 

- fixed sloppy code in user class where class and instance variables existed with same name and only the instance variables used

Please note that when using this get existing messages will not work , this is because the ru of the mock party businesses is not known to the party service. To get around this clean down your SecureMessage db and also modify the hardcoded ru in ras-frontstage.secure-message to '3b136c4b-7a14-4904-9e01-13364dd7b972' which is known to party service. Note this will need to be removed later when the clients start passing in an ru


app/services/party_service.py
```diff
@@ -0,0 +1,48 @@
...
+                              "sampleUnitType": "BI"}
+            else:
+                url = app.settings.RAS_PARTY_GET_BY_RESPONDENT.format(app.settings.RAS_PARTY_SERVICE, uuid)
+                party_data = requests.get(url, verify=False)
+                logger.debug('party get user details result => {} {} : {}'.format(party_data.status_code,
+                                                                                  party_data.reason, party_data.text))
+                party_dict = json.loads(party_data.text)
+            return Response(response=json.dumps(party_dict), status=200, mimetype="text/html")
+        except KeyError:
+            logger.debug('User ID %s not in mock party service', uuid)
```
rjweeks: We should use 'User ID {} not in mock party service'.format(uuid)

app/services/party_service.py
```diff
@@ -0,0 +1,48 @@
...
+            return Response(response="uuid not valid", status=404,
+                            mimetype="text/html")
```
rjweeks: put this on 1 line

PR #76, opened 2017-08-15
Feature/cloud foundry

rjweeks:
Use cfenv package to resolve cloud foundry environment.


run.py
```diff
@@ -3,4 +3,4 @@
 from app.application import app
 
 DEV_PORT = os.getenv('DEV_PORT', 5050)
-app.run(debug=True, host='0.0.0.0', port=int(DEV_PORT))
+app.run(debug=True, host='0.0.0.0', port=int(DEV_PORT))
```
andrewmil: should debug be true here? do we want it to be set by an environment variable?

rjweeks: run.py is not used in cloud foundry, it is the scripts that devs use to start Secure Message. So it's ok hardcoded for now.

PR #77, opened 2017-08-16
Ts 235 inform case service when new secure message sent

dcdarrell9:

- Link to Trello card.
[Card1](https://trello.com/c/Evs6fcKH) [Card2](https://trello.com/c/x7k5C9SY)

**How to review**

- Make sure the system uses the mock and the real service (can be seen with party service also).
- The tests in test_case_service_integration and test_party_service_integration are skipped, unskip them and make sure they run correctly.
- Make sure all tests run and there are no issues with secure messaging.


tests/app/test_mock_case_service.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+
+
+class CaseServiceTestCase(unittest.TestCase):
+    def test_some_function(self):
+        assert True
```
dcdarrell9: I'm not sure what/how much needs to go in here. Maybe needs another card.

tests/behavioural/features/steps/message_post.py
```diff
@@ -322,7 +322,7 @@ def step_impl_msg_is_sent(context):
     token_data['role'] = 'respondent'
     headers['Authorization'] = update_encrypted_jwt()
     context.response = app.test_client().post(url, data=json.dumps(data), headers=headers)
-
+    x=0
```
dcdarrell9: Oops.

app/services/case_service.py
```diff
@@ -0,0 +1,29 @@
...
+        """posts the data to the case service"""
+
+        json_data = {"description": "New Secure Message",
+                     "category": "SECURE_MESSAGE_SENT",
+                     "createdBy": user_name
+                     }
+
+        url = app.settings.RM_CASE_POST.format(app.settings.RM_CASE_SERVICE, case_id)
+        case_service_data = requests.post(url, json=json_data, verify=False)
+        logger.debug('case service post result => {} {} : {}'.format(case_service_data.status_code,
```
emilioward: Just a thought the logs in secure messaging could be changed to follow the format in frontstage, this could be something we could look at in the future

tests/app/test_party_integration.py
```diff
@@ -105,5 +75,5 @@ def test_calling_get_respondent_data_with_expected_test_data_returns_expected_me
                     "telephone": "1234",
                     "sampleUnitType": "BI"
                     }
-        response = sut.get_user_details("db036fd7-ce17-40c2-a8fc-932e7c228397")
-        self.assertEqual(json.loads(response.data), expected)
+        result, status_code = sut.get_user_details("db036fd7-ce17-40c2-a8fc-932e7c228397")
```
emilioward: Single quotes

app/services/case_service.py
dcdarrell9: Not sure whether we need the logging to be in the same standard as the rest of sdc, so we'll leave it for a discussion.

app/services/service_toggles.py
```diff
@@ -46,8 +48,18 @@ def get_business_details(self, ru):
...
+        super().__init__(use_mock, CaseService, CaseServiceMock, 'Case')
 
+    def store_case_event(self, case_id, user_uuid):
+        return self._service.store_case_event(case_id, user_uuid)
+
+"""party is the interface that code should use mocktoggle for the party service """
+party = Party(False)
 
+"""case_service is how code should interact with the case service"""
+case_service = Case(False)
```
andrewmil: I really don't like having to change the code to turn the mock on/off.

This should be an environment variable

app/validation/user.py
```diff
@@ -19,5 +19,5 @@ def is_respondent(self):
 
     @staticmethod
     def is_valid_user(uuid):
-        response = party.get_user_details(uuid)
-        return response.status_code == 200
+        response, status_code = party.get_user_details(uuid)
+        return status_code == 200
```
andrewmil: I don't understand the point of returning the response and status code separately if the status code is included in the response anyway. You're just returning the same thing twice

tests/app/test_draft.py
```diff
@@ -63,6 +64,8 @@ def setUp(self):
 
         self.user_internal = User('ce12b958-2a5f-44f4-a6da-861e59070a31', 'internal')
         self.user_respondent = User('0a7ad740-10d5-4ecb-b7ca-3c0384afb882', 'respondent')
+        case_service.use_mock_service()
```
andrewmil: For testing is it not better to mock the response from the party endpoint rather than using our internal mock? Would probably be simpler as well

tests/app/test_mock_party.py
```diff
@@ -142,7 +140,7 @@ def test_message_by_id_replaces_uuids(self):
                 'ru_id': 'f1a5e99c-8edf-489a-9c72-6cabe6c387fc',
                 'survey': constants.BRES_SURVEY}
 
-        encrypted_token = self._generate_encrypted_token()
+        encrypted_token = _generate_encrypted_token()
```
andrewmil: We're not using encrypted tokens anywhere in what's deployed at the moment, should the tests reflect this?

app/services/service_toggles.py
dcdarrell9: I agree.

app/services/party_service.py
```diff
@@ -13,36 +12,34 @@ class PartyService:
...
+                          "status": "",
+                          "sampleUnitType": "BI"}
+            return party_dict, 200
+        else:
+            url = app.settings.RAS_PARTY_GET_BY_RESPONDENT.format(app.settings.RAS_PARTY_SERVICE, uuid)
+            party_data = requests.get(url, verify=False)
+            logger.debug('party get user details result => {} {} : {}'.format(party_data.status_code,
+                                                                              party_data.reason, party_data.text))
+            party_dict = json.loads(party_data.text)
+            return party_dict, party_data.status_code
```
andrewmil: coverage is telling us that a lot of this code is untested. Are we wanting to do this before merging?

app/resources/messages.py
```diff
@@ -97,18 +97,58 @@ def _message_save(self, message, is_draft, draft_id):
...
-        resp.status_code = alert_status
+        # listener errors are logged but still a 201 reported
+        MessageSend._alert_listeners(message.data)
+
+        resp = jsonify({'status': '201', 'msg_id': message.data.msg_id, 'thread_id': message.data.thread_id})
+        resp.status_code = 201
         return resp
 
+    @staticmethod
+    def _alert_listeners(message):
```
andrewmil: I think emails are getting sent even when it's a respondent sending the message, is this correct?

app/resources/messages.py
```diff
@@ -97,18 +97,58 @@ def _message_save(self, message, is_draft, draft_id):
...
+            MessageSend._try_send_alert_email(message)
+            MessageSend._inform_case_service(message)
+        except Exception as e:
+            logger.error('Uncaught exception in Message.alert_listeners: {}'.format(e))
+
+    @staticmethod
+    def _try_send_alert_email(message):
+        """Send an email to recipient if appropriate"""
+        party_data = None
+        if message.msg_to[0] != constants.BRES_USER:
```
andrewmil: msg_to isn't being set to 'BRES' consistently so this piece of logic isn't always working

app/services/service_toggles.py
AndrewTorrance: I am open on this one , and can see both sides of the argument . Right now I dont think an environment variable is the correct way to go as we will never have a production use case where we need to use the mock . Its only of use for developer testing and running unit tests . So an environment variable/config is a risk for little benifit.

tests/app/test_draft.py
AndrewTorrance: For specific unit test you may be right , a mocked response might be an easier way to go , its certainly more elegant . However we also have the situation where we want to be able to run locally without having a dependency on external services . For that it its much easier to use the mocked service .

tests/app/test_mock_party.py
AndrewTorrance: Seems fair enough to me . But this change did not introduce that , so I suggest we put in a ticket to cover that work .

app/services/party_service.py
AndrewTorrance: No, I will look at coverage this morning

app/resources/messages.py
AndrewTorrance: No, its only sending when the recipient is a respondent . The _try_send_alert_email takes responsibility for determining when an email is sent .

app/resources/messages.py
AndrewTorrance: For cases where the to or from is the internal user then value 'BRES' is assumed . If this is not the case then we have much wider issues ? Can you point o cases where the internal user is not 'BRES' ?

app/resources/messages.py
andrewmil: Yeah I just followed through the code and you're right. 
I was getting a strange thing happening a few days ago where I was getting an error log telling me that gov.notify had reached max number of messages it could send even when I was just sending from frontend. Not sure what could have been triggering that

app/api_mocks/case_service_mock.py
```diff
@@ -0,0 +1,9 @@
+import logging
+
+logger = logging.getLogger(__name__)
```
andrewmil: should we be using structlog's wrap_logger like we are elsewhere?

app/resources/messages.py
andrewmil: I think Richard's PR fixed this yesterday as I'm not getting the same issues anymore

tests/app/test_mock_case_service.py
andrewmil: Case service isn't giving us any information we're actually using I think so this is probably fine

app/api_mocks/case_service_mock.py
dcdarrell9: Better to use struclog to keep consistent.

tests/app/test_party_service.py
```diff
@@ -0,0 +1,46 @@
...
+
+class PartyTestCase(unittest.TestCase):
+    """Test cases for party service"""
+
+    @staticmethod
+    def test_get_business_details_sends_request_to_remote_party_service():
+        """Test get business details sends a request"""
+        sut = PartyService()
+        business_data = PartyBusinessTestHelper(200, "OK", '{"something": "else"}')
+        requests.get = mock.Mock(name='get', return_value=business_data)
```
andrewmil: If we use the requests_mock package here we wouldn't need to define the helper function and could do it all over one line.

The current method would make it tricky to carry out tests which hit two separate requests objects as well

app/services/service_toggles.py
dcdarrell9: For now are we leaving this as is then have a discussion on how to do it? @andrewmil @AndrewTorrance

dcdarrell9:
@AndrewTorrance @rhiGriff I've removed the tests for the logging in test_service_mock_toggle for now, we can keep it out or add it back later.

app/services/case_service.py
rhiGriff: Can we update the logs to be consistent with the rest of the project now please?

app/resources/messages.py
```diff
@@ -97,21 +99,61 @@ def _message_save(self, message, is_draft, draft_id):
...
+            party_data, status_code = party.get_user_details(message.msg_to[0])  # todo avoid 2 lookups (see validate)
+            if status_code == 200:
+                if 'emailAddress' in party_data and len(party_data['emailAddress'].strip()) > 0:
+                    recipient_email = party_data['emailAddress'].strip()
+                    alert_method = AlertViaLogging() if settings.NOTIFY_VIA_LOGGING == '1' else AlertViaGovNotify()
+                    alert_user = AlertUser(alert_method)
+                    alert_user.send(recipient_email, message.msg_id)
+                else:
+                    logger.error('UserId {0} does not have an emailAddress specified'.format(message.msg_to[0]))
```
dcdarrell9: @rhiGriff @AndrewTorrance Does this log need to change?

app/services/case_service.py
dcdarrell9: 👍  Needs to be changed

app/api_mocks/case_service_mock.py
rjweeks: This needs to use structlogs wrap_logger. Or delete the log import as logger is not used.

app/resources/messages.py
```diff
@@ -97,21 +99,61 @@ def _message_save(self, message, is_draft, draft_id):
...
 
+    @staticmethod
+    def _alert_listeners(message):
+        """used to alert user and case service once messages have been saved"""
+        try:
+            MessageSend._try_send_alert_email(message)
+            MessageSend._inform_case_service(message)
+        except Exception as e:
+            logger.error('Uncaught exception in Message.alert_listeners: {}'.format(e))
+
```
rjweeks: log format

app/resources/messages.py
```diff
@@ -97,21 +99,61 @@ def _message_save(self, message, is_draft, draft_id):
 
         if is_draft is True:
             Modifier().del_draft(draft_id)
-        return MessageSend._alert_recipients(message.data.msg_id, message.data.thread_id)
 
-    @staticmethod
-    def _alert_recipients(msg_id, thread_id):
-        """used to alert user once messages have been saved"""
-        recipient_email = settings.NOTIFICATION_DEV_EMAIL  # TODO change this when know more about party service
-
```
rjweeks: don't comment inline

PR #78, opened 2017-08-16
Changes to handle 404 from support services.

rjweeks:
These are changes to handle 404's from other services better. Supporting change in Frontstage PR.

### DIFF
--- a/app/services/party_service.py
+++ b/app/services/party_service.py
@@ -35,13 +35,14 @@ class PartyService:
                               "telephone": "",
                               "status": "",
                               "sampleUnitType": "BI"}
+                return Response(response=json.dumps(party_dict), status=200, mimetype="text/html")
             else:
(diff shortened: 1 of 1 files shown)


app/services/party_service.py
```diff
@@ -41,7 +41,7 @@ def get_user_details(uuid):
                 logger.debug('party get user details result => {} {} : {}'.format(party_data.status_code,
                                                                                   party_data.reason, party_data.text))
                 party_dict = json.loads(party_data.text)
-            return Response(response=json.dumps(party_dict), status=200, mimetype="text/html")
+            return Response(response=json.dumps(party_dict), status=party_data.status_code, mimetype="text/html")
```
andrewmil: `party_data` is not defined if `uuid == constants.BRES_USER` is true

rjweeks: Good spot.

PR #80, opened 2017-08-17
configured output to logs instead of gov.notify

emilioward:

Link to Trello card 
Related to - https://trello.com/c/v9J6DtQe 

This PR adds in an ENV variable so that messaged don't get sent to gov notify and instead get outputted to the logs.

**How to review**

Send a message and check the logs that the output is there. 
Setting the ENV to `0` will send to gov.notify and setting it to `1` will output the message to the logs and not output to gov.notify.


app/common/alerts.py
```diff
@@ -20,17 +20,28 @@ def send(email, reference):
...
-            logger.exception(e)
-        finally:
-            return 201, 'OK'
+        if self.NOTIFY_VIA_LOGGING == '1':
+            try:
+                logger.info('email details {}, {}'.format(email, reference))
+            except BaseException as e:
+                logger.exception(e)
+            finally:
+                return 201, 'OK'
```
andrewmil: This part is untested, probably want to add a test which covers this chunk

app/common/alerts.py
```diff
@@ -20,17 +20,28 @@ def send(email, reference):
 
 
 class AlertUser:
-    """Alert User"""
+    """Alert goes via gov notify (0) or via logs (1)"""
+
+    NOTIFY_VIA_LOGGING = settings.NOTIFY_VIA_LOGGING
```
andrewmil: A more standard way of setting this is probably having this in an `__init__` function

app/common/alerts.py
```diff
@@ -20,17 +20,28 @@ def send(email, reference):
...
+            try:
+                logger.info('email details {}, {}'.format(email, reference))
+            except BaseException as e:
+                logger.exception(e)
+            finally:
+                return 201, 'OK'
+        else:
+            try:
+                self.alert_method.send(email, reference)
+            except BaseException as e:
```
andrewmil: This was already here I think so probably not a change for this PR but in general you don't want to be catching such a broad class of exception. Should be more specific to what's happening here

PR #93, opened 2017-08-31
Feature/thread msg list

rjweeks:
Paginate list of messages for thread/{id} end point.

Fix tests.


andrewmil:
I agree with tej the tests shouldn't be dependant on any external services.
and the gov.notify url is probably the worst one we could be hitting

rjweeks:
Travis is just rebuilding, it should use log notify this time.

PR #95, opened 2017-09-04
role is no longer required

andrewmil:

https://trello.com/c/caqTLg2d/566-bug-if-a-bres-user-token-is-encypted-with-an-empty-role-then-the-user-can-still-get-messages

Commented out the authentication around role so it is not a required field in the JWT anymore

**How to review**

To test this I've been removing the role in frontstage in `frontstage/jwt.py line 22`


andrewmil:
Just spotted a bug with this way of doing things will try and fix now

andrewmil:
Exception is getting raised for empty role string
The existing behaviour is that secure-message throws a 400 with
`description="Missing role claim, role is required to access this Microservice Resource"`
when an empty role string is in the jwt.

Frontstage currently is not looking at the description for this error though

PR #97, opened 2017-09-05
Removed setting of actor to survey in draft save , also removed dupli…

AndrewTorrance:
When I set survey id to a UUID 10 tests failed . It turned out to be the use of survey as actor in save drafts ( something Andrew Millar had already found) . Changed that so that actor is not set to survey . Also found that the RetrieverTestCase was duplicated both in test_retriever and test_utilities. Deleted it from test_utilities . Deleted constants.BRES_SURVEY as no application code uses it but created test_utilities.BRES_SURVEY in order that tests pass. Set that to a UUID 

 

Link to Trello card  https://trello.com/c/dX5XGmCw


tests/app/test_utilities.py
```diff
@@ -1,248 +1 @@
...
-            'msg_id': 'd740-10d5-4ecb-b',
-            'subject': 'subject',
-            'body': 'body'
-        }
-
-        hash_object = hashlib.sha1(str(sorted(data_to_hash.items())).encode())
-        etag = hash_object.hexdigest()
-
-        self.assertEqual(generated_etag, etag)
+BRES_SURVEY ="33333333-22222-3333-4444-88dc018a1a4c"
```
andrewmil: This doesn't seem like the right place to keep this.

The location/name of the file implies it should contain tests

PR #101, opened 2017-09-07
Defect 25 information disclosed via GitHub

andrewmil:

https://trello.com/c/eIxe8JJt/564-defect-25-information-disclosed-via-github

Removed a handful of defaults for secure environment variables. 
The removed variables will have to be set on your environment to run locally, they are:
- `JWT_SECRET`
- `SECURITY_USER_NAME`
- `SECURITY_USER_PASSWORD`

`JWT_SECRET` must match `JWT_SECRET` in frontstage 
`SECURITY_USER_NAME/PASSWORD` must match the equivalent in the party and case services


andrewmil:
I've also got questions about `SM_USER_AUTHENTICATION_(PUBLIC/PRIVATE)_KEY` variables.

These are currently not used in our deployments but when we start encrypting our JWT's we'll have to separate the codes we use for testing from the production code

PR #102, opened 2017-09-11
settings developer/test environtment variables

andrewmil:

https://trello.com/c/lIKc7Jdb/557-defect-18-story-release-10-github-secrets-used-in-pre-prod

Removed protected default settings (keys etc)
Stopping deployment when required environment variables are not set
Setting environment variables for dev and test settings

**How to review**

Check that App can still run in dev mode


run.py
```diff
@@ -1,5 +1,16 @@
...
+os.environ['JWT_SECRET'] = 'vrwgLNWEffe45thh545yuby'
+os.environ['JWT_ALGORITHM'] = 'HS256'
+os.environ['SECURITY_USER_NAME'] = 'test_user'
+os.environ['SECURITY_USER_PASSWORD'] = 'test_password'
+os.environ['NOTIFY_VIA_LOGGING'] = '1'
+os.environ['NOTIFICATION_API_KEY'] = 'test_notification_api_key'
+os.environ['SERVICE_ID'] = 'test_service_id'
+os.environ['NOTIFICATION_TEMPLATE_ID'] = 'test_notification_template_id'
+os.environ['RAS_SM_PATH'] = './'
+
```
tej99: Would this not override deployment config?

andrewmil: At the moment we don't have different configuration objects in secure-message like frontstage, we're just pulling from the settings file.

So I have to set the environment variables directly for development/testing

tej99: But setting the environment variables in run.py means when secure-messaging is spun up theres no way to configure these env vars as they will always be set to that ^^

andrewmil: yes and no, the run.py script isn't run when we're deploying to cloudfoundry so these are only set when we run locally.

It's probably a good idea to let us override them like in frontstage though, I think the plan is to change this config over the the proper flask config style like in frontstage at some point

tej99: oh okay :+1:

PR #108, opened 2017-09-29
refacor party servie error handling

rhiGriff:

Link to Trello card

**How to review**

### DIFF
--- a/app/services/party_service.py
+++ b/app/services/party_service.py
@@ -24,8 +24,6 @@ class PartyService:
 
         if party_data.status_code == 200:
             party_dict = json.loads(party_data.text)
-            if type(party_dict) is list:  # if id is not a uuid returns a list not a dict
-                party_dict = {'errors': party_dict[0]}
             return party_dict, party_data.status_code
         else:
             logger.error('Party service failed', status_code=party_data.status_code, text=party_data.text, ru=ru)
--- a/app/services/party_service.py
+++ b/app/services/party_service.py
@@ -22,11 +22,14 @@ class PartyService:
                      text=party_data.text,
                      url=url)
 
-        party_dict = json.loads(party_data.text)
-        if type(party_dict) is list:  # if id is not a uuid returns a list not a dict
-            party_dict = {'errors': party_dict[0]}
-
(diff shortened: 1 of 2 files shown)


app/services/party_service.py
```diff
@@ -22,11 +22,14 @@ def get_business_details(ru):
...
-        party_dict = json.loads(party_data.text)
-        if type(party_dict) is list:  # if id is not a uuid returns a list not a dict
-            party_dict = {'errors': party_dict[0]}
-
-        return party_dict, party_data.status_code
+        if party_data.status_code == 200:
+            party_dict = json.loads(party_data.text)
+            if type(party_dict) is list:  # if id is not a uuid returns a list not a dict
+                party_dict = {'errors': party_dict[0]}
+            return party_dict, party_data.status_code
```
andrewmil: I don't understand this piece of logic. If the party service responds with a status code of 200 why would we have errors?

tej99: if type(party_dict) is list:  # if id is not a uuid returns a list not a dict
                 party_dict = {'errors': party_dict[0]}

is now redundant.

app/services/party_service.py
```diff
@@ -52,5 +55,9 @@ def get_user_details(uuid):
...
                          url=url)
 
-            party_dict = json.loads(party_data.text)
-            return party_dict, party_data.status_code
+            if party_data.status_code == 200:
+                party_dict = json.loads(party_data.text)
+                return party_dict, party_data.status_code
+            else:
+                logger.error('Party service failed', status_code=party_data.status_code, text=party_data.text, uuid=uuid)
+                return party_data.text, party_data.status_code
```
tej99: What should get returned if theres an error may need more thought. As this will now return a list for party_data.text (which is probably why there was the if statement above in the get_business_details function ^^). maybe discuss with Shane and Andrew on what the json should look like if details are not retrieved from party service.

andrewmil: This shouldn't be an issue for for frontstage calls to sm (for now anyway). How appropriate this is depends on the form of error messages party service returns

PR #115, opened 2017-10-26
[tbl111] Use pipenv for deterministic builds

benjefferies:
Pipenv enables us to lock down and guarantee what dependencies we get
when we install dependencies. This works by using the pip hashes when
running `pipenv install --dev`. The hashes are stored in a generated
file call Pipfile.lock.

We no longer need requirement files instead pipenv uses a file called
Pipfile. The documentation can be found
https://github.com/kennethreitz/pipenv.

https://trello.com/c/TNrTo9L1/733-tbl111-create-reproducible-deployments


benjefferies:
I've tried and I'm not having much luck at the moment so I'm whipping up a simple pipenv app to rule out problems with pipenv and cloudfoundry. I can see pipenv being installed and it generating a requirements file from pipfile but it errors with some issue.

benjefferies:
I've now deployed this on CF @rjweeks

benjefferies:
Currently investigating why pipenv installs dev packages https://github.com/cloudfoundry/python-buildpack/issues/88

PR #116, opened 2017-10-27
refactor for rm-notify-gateway

jwgwalton:

https://trello.com/c/kW4olZ4N/715-tbl106-p2-party-service-to-use-rm-notify-gateway

use the rm-notify-gateway instead of the gov notify python client

**How to review**

????


app/common/alerts.py
```diff
@@ -14,12 +13,20 @@ class AlertViaGovNotify:
...
+        notification = {
+            "emailAddress": email,
+            "reference": reference
+        }
+
+        url = urlparse.urljoin(settings.RM_NOTIFY_GATEWAY_URL, settings.NOTIFICATION_TEMPLATE_ID)
+
+        auth = (settings.SECURITY_USER_NAME, settings.SECURITY_USER_PASSWORD)
+
+        response = requests.post(url, auth=auth, timeout=settings.REQUESTS_POST_TIMEOUT, json=notification)
```
andrewmil: You could just use `auth=settings.BASIC_AUTH` here as we define that object already in settings

app/common/alerts.py
```diff
@@ -14,12 +13,20 @@ class AlertViaGovNotify:
...
+
+        log.info('Sent secure message email notification{} sent via RM Notify-Gateway to GOV.UK Notify.'
```
andrewmil: I'm not sure how up to date the logging is throughout secure messaging but the structlog format we're using now would have this log look like
`log.info('Sent secure message email notification sent via RM Notify-Gateway to GOV.UK Notify', message_id=response.json()["id"])

app/common/alerts.py
andrewmil: We should probably do some check on the response we're getting from the notify service
i.e.
```
if response.status_code != 200:
    (Do error stuff here)
```

jwgwalton: Just want to clarify with someone after our chat that it's not meant to fail quietly and always return a 201. Any exception is swallowed in the AlertUser class which provides a switch for gov notify and merely logging.

andrewmil: I definitely don't think that the error should just be swallowed with no other action. This email is sent after the message has been stored in the database so we need to:
- Alert ourselves that this email has not been sent (with logging/exceptions etc)
- Return a response from secure messaging which indicates that this has happened (different status code/error code maybe?)

This needs to happen on general exceptions in this chunk of code (ConnectionError etc) and when the notify service doesn't return a 200 status code

app/application.py
```diff
@@ -90,10 +90,14 @@ def _request_requires_authentication():
            and request.method != 'OPTIONS'
 
 
-@app.errorhandler(MessageSaveException)
-def handle_save_exception(error):
-    response = jsonify(error.to_dict())
-    response.status_code = error.status_code
+@app.errorhandler(Exception)
+def handle_error(error):
+    if isinstance(error, MessageSaveException) or isinstance(error, RasNotifyException):
```
benjefferies: We could split this out into different `@app.errorhandler`s

app/common/alerts.py
benjefferies: I agree with this point. We should probably check the status code returned is healthy

app/application.py
andrewmil: How does this work when one Exception is inherited from another? Does it handle the most specific one?

benjefferies: I think it works on the order they are defined so you'd want the least specific one to be at the bottom

benjefferies: http://flask.pocoo.org/docs/0.12/errorhandling/
> Changed in version 0.11: Errorhandlers are now prioritized by specificity of the exception classes they are registered for instead of the order they are registered in.

app/common/alerts.py
```diff
@@ -14,12 +14,25 @@ class AlertViaGovNotify:
...
+            response = requests.post(url, auth=settings.BASIC_AUTH, timeout=settings.REQUESTS_POST_TIMEOUT,
+                                     json=notification)
+
+            logger.info('Sent secure message email notification, via RM Notify-Gateway to GOV.UK Notify.',
+                        message_id=response.json()["id"])
+
+        except Exception as e:
+            msg = 'There was a problem sending a notification via RM Notify-Gateway to GOV.UK Notify'
+            logger.error(msg, exception=e)
+            raise RasNotifyException(msg + str(e), status_code=500)
```
tej99: So now it will return 500 which means in back-stage it will show 500 error page even though the secure message has been 'sent', instead of failing quietly like it did before. Is this correct or do we need to raise ticket for backstage to handle this scenario differently?

app/common/alerts.py
```diff
@@ -14,12 +14,25 @@ class AlertViaGovNotify:
...
+            url = urlparse.urljoin(settings.RM_NOTIFY_GATEWAY_URL, settings.NOTIFICATION_TEMPLATE_ID)
+
+            response = requests.post(url, auth=settings.BASIC_AUTH, timeout=settings.REQUESTS_POST_TIMEOUT,
+                                     json=notification)
+
+            logger.info('Sent secure message email notification, via RM Notify-Gateway to GOV.UK Notify.',
+                        message_id=response.json()["id"])
+
+        except Exception as e:
+            msg = 'There was a problem sending a notification via RM Notify-Gateway to GOV.UK Notify'
```
benjefferies: Is it worth doing this logging in the flask exception handler?

app/application.py
```diff
@@ -90,13 +90,27 @@ def _request_requires_authentication():
...
 @app.errorhandler(MessageSaveException)
 def handle_save_exception(error):
     response = jsonify(error.to_dict())
     response.status_code = error.status_code
     return response
 
 
+@app.errorhandler(Exception)
+def handle_exception(error):
+    response = jsonify({'errors': [str(error)]})
```
benjefferies: Is it safe to return the stringified exception

app/application.py
```diff
@@ -90,13 +90,27 @@ def _request_requires_authentication():
            and request.method != 'OPTIONS'
 
 
+@app.errorhandler(RasNotifyException)
+def handle_notify_exception(error):
+    response = jsonify(error.to_dict())
```
benjefferies: Is it safe to log error.to_dict? Will it leak any internal information?

app/common/alerts.py
benjefferies: We should be catching the explicit exceptions which could be thrown and handle them correctly. If the message is sent we shouldn't fall over. What exception are we handling here?

Does `msg + str(e)` give anything useful in the message?

andrewmil:
My changes have been dealt with but should deal with Ben's comments about logging

app/common/alerts.py
jwgwalton: I previously said that it should continue normally? The code was logging the error but continuing.

app/application.py
```diff
@@ -90,13 +90,23 @@ def _request_requires_authentication():
...
+@app.errorhandler(RasNotifyException)
+def handle_notify_exception(error):
     response = jsonify(error.to_dict())
     response.status_code = error.status_code
     return response
 
 
+@app.errorhandler(MessageSaveException)
+def handle_save_exception(error):
+    return jsonify(status_code=error.status_code)
```
benjefferies: Could we just get away with setting the status code instead of returning it in the body?

PR #117, opened 2017-10-31
[tbl111] Use pipenv for deterministic builds

benjefferies:
Pipenv enables us to lock down and guarantee what dependencies we get
when we install dependencies. This works by using the pip hashes when
running `pipenv install --dev`. The hashes are stored in a generated
file call Pipfile.lock.

We no longer need requirement files instead pipenv uses a file called
Pipfile. The documentation can be found
https://github.com/kennethreitz/pipenv.

* Bump cryptography library

* Use latest dev dependencies

**What is the context of this PR?**

Link to Trello card

https://trello.com/c/TNrTo9L1/733-tbl111-create-reproducible-deployments

**How to review**

1. Check out locally and follow README.md to set up with pipenv
1. `cf push` to check deployment

PR #118, opened 2017-11-03
Remove cloud foundry buildpack from manifest

jcox-dev:
#117 moved to use Python 3.6.3. The version of the custom buildpack https://github.com/cloudfoundry/python-buildpack only offered up to version 3.6.1, which caused the Jenkins build to fail. 

The microservices deployed to CF no longer require the buildpack defined in the manifest.


benjefferies:
Have you tested the `cf push` for this app @jcox-dev ?

benjefferies:
`cf push` looks good to me

PR #121, opened 2017-12-28
Replace settings with config

jcox-dev:


jcox-dev:
[Change Secure Message to use Config rather than settings](https://trello.com/c/qgRTKXxx/20-change-secure-message-to-use-config-rather-than-settings)

jcox-dev:
Probably good to go now. Will test on CF.

run.py
```diff
@@ -1,17 +1,8 @@
 #!flask/bin/python
 import os
 
-os.environ['JWT_SECRET'] = os.getenv('JWT_SECRET', 'vrwgLNWEffe45thh545yuby')
```
benjefferies: 👍

secure_message/application.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _request_requires_authentication():
+    return request.endpoint is not None and \
+           'health' not in request.endpoint \
+           and request.endpoint != 'info' \
+           and request.method != 'OPTIONS'
+
+
+def log_request():
```
benjefferies: Wonder if we need this logging as cloudfoundry does have some api logging https://docs.cloudfoundry.org/devguide/deploy-apps/streaming-logs.html

secure_message/authentication/jwe.py
```diff
@@ -10,21 +10,25 @@
...
 class Encrypter:
 
-    def __init__(self, _private_key=settings.SM_USER_AUTHENTICATION_PRIVATE_KEY,
-                 _private_key_password=settings.SM_USER_AUTHENTICATION_PRIVATE_KEY_PASSWORD,
-                 _public_key=settings.SM_USER_AUTHENTICATION_PUBLIC_KEY):
+    def __init__(self, _private_key=None, _private_key_password=None, _public_key=None):
         """ initialise encrypter with correct keys, the ability to change default keys for tests"""
 
+        if _private_key is None:
+            _private_key = current_app.config['SM_USER_AUTHENTICATION_PRIVATE_KEY']
```
benjefferies: Could be `_private_key  = _private_key  or current_app.config['SM_USER_AUTHENTICATION_PRIVATE_KEY']`

secure_message/authentication/jwt.py
```diff
@@ -2,22 +2,25 @@
...
-
-JWT_ALGORITHM = settings.SM_JWT_ALGORITHM
-JWT_SECRET = settings.SM_JWT_SECRET
 
 
 def encode(data):
     """
     Function to encode python dict data
     """
+    JWT_ALGORITHM = current_app.config['SM_JWT_ALGORITHM']
```
benjefferies: Could be lower cased too

secure_message/common/utilities.py
```diff
@@ -70,19 +68,19 @@ def paginated_list_to_json(paginated_list, page, limit, host_url, user, string_q
...
 
     if paginated_list.has_next:
         links['next'] = {
-            "href": "{0}{1}{2}{3}page={4}&limit={5}".format(host_url, endpoint, arg_joiner, string_query_args, (page + 1), limit)}
+            "href": f"{host_url}{endpoint}{arg_joiner}{string_query_args}page={page + 1}&limit={limit}"}
 
     if paginated_list.has_prev:
         links['prev'] = {
-            "href": "{0}{1}{2}{3}page={4}&limit={5}".format(host_url, endpoint, arg_joiner, string_query_args, (page - 1), limit)}
+            "href": f"{host_url}{endpoint}{arg_joiner}{string_query_args}page={page - 1}&limit={limit}"}
```
benjefferies: So much tidier with f string formatting. I think we should be using f strings across the board

secure_message/resources/drafts.py
```diff
@@ -90,7 +88,7 @@ class DraftList(Resource):
     def get():
         """Get message list with options"""
 
-        string_query_args, page, limit, ru_id, survey, cc, label, desc, ce = get_options(request.args)
+        string_query_args, page, limit, _, _, _, _, _, _ = get_options(request.args)
```
benjefferies: It would be nicer if get_options returned a dict or was removed as this looks like the only place it's used

secure_message/resources/health.py
```diff
@@ -0,0 +1,57 @@
...
+        func_list = {}
+        for rule in current_app.url_map.iter_rules():
+            if rule.endpoint != 'static':
+                func_list[rule.rule] = current_app.view_functions[rule.endpoint].__doc__
+
+        details = {'Name': current_app.config['NAME'],
+                   'Version': current_app.config['VERSION'],
+                   'SMS Log level': current_app.config['SMS_LOG_LEVEL'],
+                   'APP Log Level': current_app.config['APP_LOG_LEVEL'],
+                   'Database URL': current_app.config['SQLALCHEMY_DATABASE_URI'],
```
benjefferies: Should we be exposing Database URL in health checks as it contains the DB credentials

secure_message/resources/drafts.py
jcox-dev: Agreed. It is used elsewhere but also probably beyond the scope of this PR to refactor the utility.

secure_message/resources/health.py
jcox-dev: Again, beyond the scope of this PR but good to flag for consideration.

secure_message/application.py
jcox-dev: I also considered stripping this out but would require a discussion separate to this PR.

secure_message/authentication/jwt.py
rhiGriff: Does this affect the Jenkins config? If so we don't want to change the case as it will affect Jenkins

PR #123, opened 2018-01-10
Fix dockerfile

insacuri:
The f-string notation didn't come until python 3.6 which is why the image was updated.
The dockerfile didn't have any environment variables set so now it'll work out of the
box in accordance with the readme.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,15 +4,8 @@ ENV BUILD_PACKAGES="build-essential python3-dev python3-pip libpq-dev"
 WORKDIR /app
 COPY . /app
 EXPOSE 5050
+
 RUN apt-get update && apt-get install -y $RUNTIME_PACKAGES $BUILD_PACKAGES
 RUN pip3 install pipenv==8.3.1 && pipenv install --system --deploy
-
-ENV JWT_SECRET testsecret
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,10 +1,18 @@
...
+FROM python:3.6
 ENV RUNTIME_PACKAGES="python3"
 ENV BUILD_PACKAGES="build-essential python3-dev python3-pip libpq-dev"
 WORKDIR /app
 COPY . /app
 EXPOSE 5050
 RUN apt-get update && apt-get install -y $RUNTIME_PACKAGES $BUILD_PACKAGES
 RUN pip3 install pipenv==8.3.1 && pipenv install --system --deploy
+
+ENV JWT_SECRET testsecret
```
andrewmil: Is including environment variables like this in the `Dockerfile` the correct way to supply them? I can't see us doing it in the other repos

insacuri: How do other repos run in docker then?  Are these environment variables in the docker-compose.yml file or does everyone run docker by using `docker run --env env1......`?

insacuri:
I've moved the environment variables into the docker-compose.yml file.  Because this file is only used by this single repository, it shouldn't affect ras-compose

PR #124, opened 2018-01-16
Feature/flake8 pylint changes

AndrewTorrance:
There were numerous flake8 and pylint errors being reported on the project which meant that it was almost pointless running flake8 and pylint. 

Link to Trello card
https://trello.com/c/bySNwAvC

Mostly the changes are adding spaces , aligning text etc and adding a .flake8 file that set the max line length to 160 .  One or two code changes :- changed using 'if len(sequence)>0' and replaced with 'if sequence'. Also added a property to the service-toggle class to remove the need for additional boolean. Also marked some errors as ti be ignored by flake8 and pylint . Some of these may benefit from being revisited at some point.

**How to review**


AndrewTorrance:
[Flake8 issues on secure-messages (2)](https://trello.com/c/bySNwAvC/600-flake8-issues-on-secure-messages-2)

.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+ignore = E402
+max-line-length=160
```
insacuri: 160? Seems a bit high

andrewmil: We use 120 across most of the other repos

secure_message/resources/threads.py
```diff
@@ -17,12 +17,13 @@ class ThreadById(Resource):
     @staticmethod
     def get(thread_id):
         """Get messages by thread id"""
-        string_query_args, page, limit, ru_id, survey, cc, label, desc, ce = get_options(request.args)
+        string_query_args, page, limit, _, _1, _2, _3, _4, _5 = get_options(request.args)  # NOQA TODO - Named Tuple
```
insacuri: Are these args documented anywhere?  I think they are but wanted to double check because the variable names acted like documentation (even though they probably fail pylint for not being used)

secure_message/resources/messages.py
```diff
@@ -114,17 +114,17 @@ def _alert_listeners(message):
         try:
             MessageSend._try_send_alert_email(message)
             MessageSend._inform_case_service(message)
-        except Exception as e:  # NOQA
+        except Exception as e:  # NOQA pylint:disable=broad-except
```
insacuri: Won't affect the mergability of this PR, but do you know if this super broad exception is legit or just laziness with exception typing?

.flake8
AndrewTorrance: Copied 160 from sdx collect . No probs changing , but with modern screens that fits ok.

secure_message/resources/messages.py
AndrewTorrance: The try send alert message sends via gov.notify and case service via a post. The idea here was that no matter what went wrong then informing other services of the new message should not cause an error on the message save.Hence the catch all .

tests/behavioural/features/steps/etag.py
```diff
@@ -12,14 +12,14 @@ def step_impl_etag_should_be_sent_with_draft(context):
...
+def step_impl_etag_is_requested_with_empty_value(context):
     """specify an etag but do not give it a value"""
     context.bdd_helper.headers["Etag"] = ""
 
 
 @given("an etag is requested with a value of '{etag}'")
 @when("an etag is requested with a value of '{etag}'")
-def step_impl_etag_is_requested(context, etag):
+def step_impl_etag_is_requested_with_specific_value(context, etag):
     """specify an etag ofa specificvalue"""
```
insacuri: Mind sorting out the spacing here while you're there?

gunicorn.py
```diff
@@ -179,34 +179,42 @@
...
 
+
 def worker_int(worker):
     worker.log.info("worker received INT or QUIT signal")
 
-    ## get traceback info
-    import threading, sys, traceback
+    # get traceback info
+    import threading
+    import sys
```
gemmai95: Pep8 normally requires imports to be in alphabetical order

secure_message/application.py
```diff
@@ -75,10 +75,7 @@ def handle_exception(error):  # NOQA
 
 
 def _request_requires_authentication():
-    return request.endpoint is not None and \
-           'health' not in request.endpoint \
-           and request.endpoint != 'info' \
-           and request.method != 'OPTIONS'
+    return request.endpoint is not None and 'health' not in request.endpoint and request.endpoint != 'info'and request.method != 'OPTIONS'
```
dcdarrell9: Needs a space in between `'info'and` and also 140 characters so would be over the 120 mark

secure_message/logger_config.py
```diff
@@ -26,7 +26,7 @@ def logger_initial_config(service_name=None,
     except ValueError:
         indent = None
 
-    def add_service(logger, method_name, event_dict):
+    def add_service(_1, _2, event_dict):
```
dcdarrell9: Would it better not to have `_1, _2` and have `_, _` ?

secure_message/resources/threads.py
dcdarrell9: Same here with `_` unless there's a specific name for reference, may as well have it as `_`

tests/behavioural/features/steps/to_field.py
```diff
@@ -63,4 +63,4 @@ def step_impl_the_msg_to_is_set_to_internal_as_string_not_array(context):
 def step_impl_retrieved_msg_to_is_as_saved(context):
     """validate that the message to field in the response is teh same aswas saved"""
```
dcdarrell9: `teh same aswas` 🥇

tests/behavioural/features/steps/etag.py
AndrewTorrance: done

secure_message/logger_config.py
AndrewTorrance: No as that causes an error , the _X syntax allows multiple unused variables. The better answer would be to convert to a named tuple

secure_message/resources/threads.py
AndrewTorrance: see above

tests/behavioural/features/steps/to_field.py
AndrewTorrance: Done

.flake8
AndrewTorrance: leaving at 160 until there is agreement cross teams , we have 120 in RAS , 150 in EQ and 160 in SDX.

secure_message/logger_config.py
AndrewTorrance: Have changed to a named tuple approach , makes it cleaner .

PR #129, opened 2018-01-23
enumeration added and replaced string with the related enum

atrozzo:

Link to Trello card 
https://trello.com/c/DfbiXbnl/27-sm-api-use-enumeration-for-possible-events-instead-of-current-strings-2

Replaced all the Strings that represent change of event in the message status, eg : Sent , Draft_Saved, Read , with an enumeration value that represent that String. 

**How to review**


secure_message/repository/modifier.py
```diff
@@ -3,6 +3,9 @@
 from flask import jsonify
 from structlog import wrap_logger
 from werkzeug.exceptions import InternalServerError
+
+from secure_message.common import events
```
AndrewTorrance: from secure_message.common import events is not needed as we are importing Events ?

rhiGriff:
Can we update the branch as well please?

warrenbailey:
Can you update the PR description too?

PR #131, opened 2018-01-26
Fix startup config errors

### LINKED ISSUE #130: Application fails to start using Dev and Test configs
The application fails to start unless *all* configuration requirements are set in the environment, even though under a dev and test environment the application should be using sensible defaults.

The behaviour is caused by [this](https://github.com/ONSdigital/ras-secure-message/blob/acb3d5f96eef0ecc4a1ac768bb3d9bc5b889e61c/config.py#L123) for loop.

Removing the for loop will allow the app to properly start up, however there should still be a check in a prod environment that those env vars are set, to ensure the app is correctly configured on startup.

JamesGardiner:

Fixes #130. Removes a for loop at the bottom of the config file that checked whether non default vars were set. This would always fail when running dev or test, as these environments expect the `TestConfig` and `DevConfig` config classes to be used.

This PR also removes workarounds put in place to mitigate this bug.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -7,12 +7,7 @@ if __name__ == "__main__":
     os.environ['APP_SETTINGS'] = 'TestConfig'
 
     # NB: the following are set to avoid MissingEnvironmentVariable being raised in config.py
-    os.environ['JWT_SECRET'] = 'testsecret'
-    os.environ['SECURITY_USER_NAME'] = 'admin'
     os.environ['SECURITY_USER_PASSWORD'] = 'secret'
-    os.environ['NOTIFICATION_API_KEY'] = 'test_notification_api_key'
-    os.environ['NOTIFICATION_TEMPLATE_ID'] = 'test_notification_template_id'
-    os.environ['SERVICE_ID'] = 'test_service_id'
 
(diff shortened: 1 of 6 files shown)


run_tests.py
```diff
@@ -7,12 +7,7 @@
     os.environ['APP_SETTINGS'] = 'TestConfig'
 
     # NB: the following are set to avoid MissingEnvironmentVariable being raised in config.py
-    os.environ['JWT_SECRET'] = 'testsecret'
```
insacuri: Not a problem per se, but this block was useful when getting this running as this repo doesn't hint at any sensible defaults for these.  Could you add something somewhere in this repo to let new devs get started (maybe in the readme)?

secure_message/authentication/jwt.py
```diff
@@ -12,7 +12,7 @@ def encode(data):
     """
     jwt_algorithm = current_app.config['SM_JWT_ALGORITHM']
     jwt_secret = current_app.config['SM_JWT_SECRET']
-
+    print(data)
```
necrophonic: Doesn't look like this should be here?

PR #134, opened 2018-01-29
implemented caching mechanism for uiid and ru already called for each…

atrozzo:
… request.

Link to Trello card : https://trello.com/c/EJbRooMU/29-use-local-per-request-cache-in-classes-that-interface-to-external-services-3

In party service I have stored the the uuid and the ru for each request, so that if the call for that specific id has been done, the stored party_data will be returned without calling the real service. 

**How to review**

### DIFF
--- a/secure_message/services/party_service.py
+++ b/secure_message/services/party_service.py
@@ -16,11 +16,6 @@ class PartyService:
 
     @staticmethod
     def get_url(api_param, code):
-        """
-        :param api_param: is the key configuration that represents part of the URI.
-        :param code: is the code to use in the url ( uuid or ru )
-        :return: a formatted url
-        """
         try:
             current_app.config[api_param].format(current_app.config['RAS_PARTY_SERVICE'], code)
         except KeyError:
@@ -45,10 +40,6 @@ class PartyService:
         return party_data.text, party_data.status_code
 
     def get_user_details(self, uuid):
-        """
-        :param uuid:
-        :return: Return user details , unless user is Bres in which case return constant data
-        """
         if uuid == constants.BRES_USER:
             party_dict = {"id": constants.BRES_USER,
                           "firstName": "BRES",
--- a/secure_message/services/party_service.py
+++ b/secure_message/services/party_service.py
@@ -17,7 +17,7 @@ class PartyService:
     @staticmethod
     def get_url(api_param, code):
         try:
-            current_app.config[api_param].format(current_app.config['RAS_PARTY_SERVICE'], code)
+            return current_app.config[api_param].format(current_app.config['RAS_PARTY_SERVICE'], code)
         except KeyError:
             raise KeyError(f"{api_param} not present")
 
@@ -28,7 +28,7 @@ class PartyService:
             logger.info("Party Service: retrieve party data using", ru=ru)
             party_data = requests.get(PartyService.get_url('RAS_PARTY_GET_BY_BUSINESS', ru),
                                       auth=current_app.config['BASIC_AUTH'], verify=False)
-            self.__business_details_cache.update(ru, party_data)
+            self.__business_details_cache[ru] = party_data
         else:
             party_data = self.__business_details_cache.get(ru)
 
@@ -54,7 +54,7 @@ class PartyService:
             logger.info("Party Service: retrieve party data using", uuid=uuid)
             party_data = requests.get(PartyService.get_url('RAS_PARTY_GET_BY_RESPONDENT', uuid),
                                       auth=current_app.config['BASIC_AUTH'], verify=False)
-            self.__users_cache.update(uuid, party_data)
+            self.__users_cache[uuid] = party_data
         else:
             party_data = self.__users_cache.get(uuid)
 
--- a/secure_message/services/party_service.py
+++ b/secure_message/services/party_service.py
@@ -10,18 +10,30 @@ logger = wrap_logger(logging.getLogger(__name__))
 
 
 class PartyService:
+
+    users_cache = dict()
+    business_details_cache = dict()
+
     @staticmethod
-    def get_business_details(ru):
+    def get_url(api_param, code):
+        """
+        :param api_param:  is the key configuration that that represent part or the URI .
+        :param code: is the code to use in the url ( uuid or ru )
+        :return: a formatted url
+        """
+        if current_app.config[api_param] is None or current_app.config['RAS_PARTY_SERVICE'] is None:
+            raise KeyError("%s not present" % api_param)
+        return current_app.config[api_param].format(current_app.config['RAS_PARTY_SERVICE'], code)
+
+    def get_business_details(self, ru):
         """Retrieves the business details from the party service"""
 
-        url = current_app.config['RAS_PARTY_GET_BY_BUSINESS'].format(current_app.config['RAS_PARTY_SERVICE'], ru)
-        party_data = requests.get(url, auth=current_app.config['BASIC_AUTH'], verify=False)
-
-        logger.debug('Party service get business details result',
-                     status_code=party_data.status_code,
-                     reason=party_data.reason,
-                     text=party_data.text,
-                     url=url)
+        if ru not in self.business_details_cache:
+            party_data = requests.get(PartyService.get_url('RAS_PARTY_GET_BY_BUSINESS', ru),
+                                      auth=current_app.config['BASIC_AUTH'], verify=False)
+            self.users_cache.update(ru, party_data)
+        else:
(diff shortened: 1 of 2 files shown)


secure_message/services/party_service.py
```diff
@@ -10,18 +10,30 @@
...
 
-        logger.debug('Party service get business details result',
-                     status_code=party_data.status_code,
-                     reason=party_data.reason,
-                     text=party_data.text,
-                     url=url)
+        if ru not in self.business_details_cache:
+            party_data = requests.get(PartyService.get_url('RAS_PARTY_GET_BY_BUSINESS', ru),
+                                      auth=current_app.config['BASIC_AUTH'], verify=False)
+            self.users_cache.update(ru, party_data)
```
jcox-dev: Would expect to be using the business details cache Inside the `get_business_details` method: `users_cache` -> `business_details_cache`. (Same again a couple of lines below).

secure_message/services/party_service.py
```diff
@@ -30,9 +42,11 @@ def get_business_details(ru):
         logger.error('Party service failed', status_code=party_data.status_code, text=party_data.text, ru=ru)
         return party_data.text, party_data.status_code
 
-    @staticmethod
-    def get_user_details(uuid):
-        """Return user details , unless user is Bres in which case return constant data"""
+    def get_user_details(self, uuid):
+        """
+        :param uuid:
+        :return: Return user details , unless user is Bres in which case return constant data
```
jcox-dev: Extra space after "details" and before ",".

secure_message/services/party_service.py
```diff
@@ -41,18 +55,16 @@ def get_user_details(uuid):
...
-                                  auth=current_app.config['BASIC_AUTH'],
-                                  verify=False)
-
-        logger.debug('Party get user details result',
-                     status_code=party_data.status_code,
-                     reason=party_data.reason,
-                     text=party_data.text,
-                     url=url)
+        if uuid not in self.users_cache:
+            logger.info("Executing call to the  party service ")
```
jcox-dev: Might be worth adding the `uuid` to the log here.

secure_message/services/party_service.py
```diff
@@ -10,18 +10,30 @@
 
 
 class PartyService:
+
+    users_cache = dict()
```
jcox-dev: Should these be prefixed with underscores (i.e.: `_users_cache`) if they're specific only to this class?

secure_message/services/party_service.py
```diff
@@ -10,18 +10,30 @@
...
+    business_details_cache = dict()
+
     @staticmethod
-    def get_business_details(ru):
-        """Retrieves the business details from the party service"""
+    def get_url(api_param, code):
+        """
+        :param api_param:  is the key configuration that that represent part or the URI .
```
JamesGardiner: Typo - two 'that's

secure_message/services/party_service.py
```diff
@@ -10,18 +10,30 @@
...
+        :param code: is the code to use in the url ( uuid or ru )
+        :return: a formatted url
+        """
+        if current_app.config[api_param] is None or current_app.config['RAS_PARTY_SERVICE'] is None:
+            raise KeyError("%s not present" % api_param)
```
JamesGardiner: This would be more Pythonic:

```Python
try:
    current_app.config[api_param].format(current_app.config['RAS_PARTY_SERVICE'], code)
except KeyError:
	raise KeyError(f"{api_param} not present")
```

secure_message/services/party_service.py
necrophonic: +1 Definitely need contexts in logging

secure_message/services/party_service.py
```diff
@@ -41,18 +55,16 @@ def get_user_details(uuid):
                           "telephone": "",
                           "status": "",
                           "sampleUnitType": "BI"}
+            logger.info("UUID is BRES")
```
necrophonic: No useful information in this log message

necrophonic:
Try to avoid commits with messages like "code review fixes" - every commit message should stand on it's own in the future away from the context of a PR. Squash commits if required.

secure_message/services/party_service.py
```diff
@@ -17,13 +17,14 @@ class PartyService:
     @staticmethod
     def get_url(api_param, code):
         """
-        :param api_param:  is the key configuration that that represent part or the URI .
+        :param api_param:  is the key configuration that represent part or the URI .
```
jcox-dev: Should this read "is the key configuration that represent**s** part **of** the URI"?

Extra space between "URI" and "."

secure_message/services/party_service.py
```diff
@@ -10,18 +10,31 @@
...
+    __users_cache = dict()
+    __business_details_cache = dict()
+
     @staticmethod
-    def get_business_details(ru):
-        """Retrieves the business details from the party service"""
+    def get_url(api_param, code):
+        """
+        :param api_param:  is the key configuration that represent part or the URI .
+        :param code: is the code to use in the url ( uuid or ru )
```
jcox-dev: Extra space between parentheses -> (uuid or ru)

secure_message/services/party_service.py
```diff
@@ -10,18 +10,31 @@
...
+
+    __users_cache = dict()
+    __business_details_cache = dict()
+
     @staticmethod
-    def get_business_details(ru):
-        """Retrieves the business details from the party service"""
+    def get_url(api_param, code):
+        """
+        :param api_param:  is the key configuration that represent part or the URI .
```
jcox-dev: To stay consistent with the rest of the docstrings, it's probably worth dropping the space and "." from the end of this line.

secure_message/services/party_service.py
jcox-dev: ^ still present

secure_message/services/party_service.py
atrozzo: I wonder if this is the pep8 . Imean comments are not code !

jcox-dev: No, not PEP8. Changes requested for consistent grammar.

secure_message/services/party_service.py
```diff
@@ -59,7 +60,7 @@ def get_user_details(self, uuid):
             return party_dict, 200
 
         if uuid not in self.__users_cache:
-            logger.info("Executing call to the  party service ")
+            logger.info("Party Service: retrieve party data using", uuid=uuid)
```
jcox-dev: We can drop the "using" here as the `uuid` is now separate from this statement.

secure_message/services/party_service.py
```diff
@@ -30,6 +30,7 @@ def get_business_details(self, ru):
         """Retrieves the business details from the party service"""
 
         if ru not in self.__business_details_cache:
+            logger.info("Party Service: retrieve party data using", ru=ru)
```
jcox-dev: We can drop the "using" here as the `ru` is now separate from this statement.

necrophonic:
Changelog missing/not updated

secure_message/services/party_service.py
```diff
@@ -10,18 +10,27 @@
 
 
 class PartyService:
+
+    __users_cache = dict()
```
JamesGardiner: Generally the syntax for such operations is `my_var = {}`. This is both faster and generally more widely used. See this for an analysis of the performance https://doughellmann.com/blog/2012/11/12/the-performance-impact-of-using-dict-instead-of-in-cpython-2-7-2/ (it is for Python 2.7.2 so a little outdated now)

JamesGardiner: What is the reason for making these private? Generally the only use case for dunderscores is the following:

"If your class is intended to be subclassed, and you have attributes that you do not want subclasses to use, consider naming them with double leading underscores and no trailing underscores. This invokes Python's name mangling algorithm, where the name of the class is mangled into the attribute name. This helps avoid attribute name collisions should subclasses inadvertently contain attributes with the same name." - see [pep8](https://www.python.org/dev/peps/pep-0008/#designing-for-inheritance) for more details

secure_message/services/party_service.py
atrozzo: nope sorry I am not going to change it.

warrenbailey:
@necrophonic at the minute none of the RAS/RM repositories have CHANGELOG files. I think there should be a tasks created to add them. It shouldn't fall to this PR to add it.

warrenbailey:
I'm happy with this PR once @JamesGardiner comments have been address, once the log message requested by @jcox-dev is updated and once the tests pass.

The changelog will be added in a separate PR.

PR #135, opened 2018-01-30
Feature/db changes for multiple surveys

AndrewTorrance:
Changes to db to support both a message to a group and identify the specific user within that group . c.f.  internal_sent_audit which is intended to be removed .Also fixed a bug where some tests were looking for a real party service not a mock.

Trello Card
https://trello.com/c/3WZq4frf

Created a new table called actors which is intended to have one row per recipient in a conversation. Identifies the to and from actor and if the from actor is internal. The intention is to support the ability to identify both the group ( via survey id) and optionally the id of the internal user . Optionally because the respondent may not know the internal user to send to.
Currently only the sent_from_internal flag is returned in the json . Future stories will read the additional fields .

Note , this performs db changes locally . Alembic changes for roll out to other environments will be needed in a separate ticket.


secure_message/repository/saver.py
```diff
@@ -64,3 +64,16 @@ def save_msg_audit(msg_id, msg_urn, session=db.session):
...
+    def save_msg_actors(msg_id, from_actor, to_actor, sent_from_internal, session=db.session):
+        """Save actor information data to database"""
+        actor = Actors(msg_id, from_actor, to_actor, sent_from_internal)
+
+        try:
+            session.add(actor)
+            session.commit()
+        except Exception as e:
+            session.rollback()
+            logger.error('Message actor save failed', error=e)
```
rhiGriff: This should be a logger.exception, you won't need the as e or to pass the e as an argument.

tests/app/test_saver.py
```diff
@@ -82,6 +82,46 @@ def test_save_msg_audit_raises_message_save_exception_on_db_error(self):
...
+        mock_session.commit.side_effect = Exception("Not Saved")
+        message = {'msg_id': 'MsgId', 'msg_to': 'Tej', 'msg_from': 'someone'}
+        with self.app.app_context():
+            with current_app.test_request_context():
+                with self.assertRaises(MessageSaveException):
+                    Saver().save_msg_actors(message['msg_id'], message['msg_to'], message['msg_from'], False, mock_session)
+
+    def test_save_actors_saves_expected_information(self):
+        """Simple read write validation of actor table"""
+        pass
```
necrophonic: Shouldn't be here?

necrophonic:
Changelog missing/not updated

secure_message/resources/messages.py
```diff
@@ -89,13 +89,13 @@ def _message_save(message, is_draft, draft_id):
         save = Saver()
         save.save_message(message.data)
         save.save_msg_event(message.data.msg_id, 'Sent')
+        save.save_msg_actors(message.data.msg_id, message.data.msg_from, message.data.msg_to[0], g.user.is_internal)
```
insacuri: Does there need to be any validation to make sure none of these are empty or does it not matter?  I **think** it doesn't matter but I thought I'd ask the question

tests/behavioural/features/steps/common.py
```diff
@@ -54,5 +54,12 @@ def step_impl_reuse_the_nth_sent_message(context, message_index):
...
     """ validate that the correct number of messages was returned"""
-    nose.tools.assert_equal(int(message_count), len(
-        context.bdd_helper.messages_responses_data[0]['messages']))
+    nose.tools.assert_equal(int(message_count), len(context.bdd_helper.messages_responses_data[0]['messages']))
+
+
+@given("debug step")
+@when("debug step")
+@then("debug step")
+def step_impl_debug_step(context):
```
insacuri: Does this still need to be here?

secure_message/resources/messages.py
AndrewTorrance: it does matter but validation takes place on entry before getting here

tests/behavioural/features/steps/common.py
AndrewTorrance: Can be dropped , but very useful when debugging features . I have written it and taken it out about 5 times before this . Thought I would leave it in in case others find it useful

tests/app/test_saver.py
AndrewTorrance: I know that there is an argument for not doing this sort of tests if you view it as testing the db. However It does test the __init__ and the set from domain model code in the Actors class .

secure_message/repository/saver.py
AndrewTorrance: Good call . I copied the existing code , its wrong in the except blocks for all the tables .

secure_message/repository/database.py
```diff
@@ -57,7 +58,7 @@ def set_from_domain_model(self, domain_model):
         self.survey = domain_model.survey
         self.collection_exercise = domain_model.collection_exercise
 
-    def serialize(self, user, body_summary=False):  # pylint:disable=too-complex
+    def serialize(self, user, body_summary=False):  # pylint:disable=too-complex, too-many-branches
```
warrenbailey: hmm .. not sure adding to the pylint is correct thing to do here. I'm actually surprised we have too-complex disabled.

This method looks fair to complicated tbh and should be refactored.

secure_message/repository/saver.py
warrenbailey: Why we catching exception and not the relevant sqlalchemy exception?

AndrewTorrance: Good call, I will tidy them up . Its the same on all of the tables not just the new one

secure_message/repository/database.py
AndrewTorrance: Fair comment. Will do . It was added when we had several hundred pylint errors and we wanted to get to zero.

tests/behavioural/features/steps/common.py
AndrewTorrance: I think if we paid for pycharm it may allow debugging of features , but since we don't we can add a debug step in a feature file and breakpoint there . This is needed because the bdd tests reuse steps in many features , so break pointing a step is a pain .

secure_message/resources/messages.py
JamesGardiner: When you say takes place on entry, do you mean in the UI?

tests/app/test_saver.py
```diff
@@ -75,13 +76,53 @@ def test_save_msg_audit_does_not_raise_exception_on_successful_save(self):
...
+                    {'msg_id': 'MsgId2', 'msg_to': '', 'msg_from': 'someone', 'sent_from_internal': False},
+                    {'msg_id': 'MsgId3', 'msg_to': 'aperson', 'msg_from': '', 'sent_from_internal': False}]
+        with self.app.app_context():
+            with current_app.test_request_context():
+                for message in messages:
+                    Saver().save_message(SecureMessage(message['msg_id']))
+                    Saver().save_msg_actors(message['msg_id'], message['msg_from'], message['msg_to'], message['sent_from_internal'])
+        with self.engine.connect() as con:
+            for message in messages:
+                query = "SELECT * FROM securemessage.actors WHERE msg_id='{}'".format(message['msg_id'])
```
JamesGardiner: f string would be more succinct

secure_message/repository/database.py
```diff
@@ -157,6 +170,37 @@ def serialize(self):
...
 
 
+class Actors(db.Model):
+    """Label Assignment table model"""
+    __tablename__ = "actors"
+
+    id = Column("id", Integer(), primary_key=True)
+    msg_id = Column('msg_id', String(constants.MAX_MSG_ID_LEN + 1), ForeignKey('secure_message.msg_id'), index=True)
+    from_actor = Column('from_actor', String(constants.MAX_STATUS_ACTOR_LEN + 1))
+    to_actor = Column('to_actor', String(constants.MAX_STATUS_ACTOR_LEN + 1))
```
necrophonic: Are these always `+1`? Would it be advantageous to create the constant with the extra in the first place?

secure_message/repository/database.py
```diff
@@ -157,6 +170,37 @@ def serialize(self):
...
+
+    def set_from_domain_model(self, msg_id, from_actor, to_actor, sent_from_internal):
+        """Set actors values"""
+        self.msg_id = msg_id
+        self.from_actor = from_actor
+        self.to_actor = to_actor
+        self.sent_from_internal = sent_from_internal
+
+    @property
+    def serialize(self):
```
necrophonic: Does serialize actually serialize, or just return data suitable for serialization?

The name suggests the former; the comment suggests the latter

secure_message/repository/saver.py
```diff
@@ -19,9 +20,9 @@ def save_message(domain_message, session=db.session):
         try:
             session.add(db_message)
             session.commit()
-        except Exception as e:
+        except SQLAlchemyError:
             session.rollback()
-            logger.error('Message save failed', error=e)
+            logger.exception('Secure message save failed')
```
necrophonic: Any kind of id or context that can be added to help issue resolution?

secure_message/repository/saver.py
```diff
@@ -33,9 +34,9 @@ def save_msg_status(actor, msg_id, label, session=db.session):
         try:
             session.add(db_status_to)
             session.commit()
-        except Exception as e:
+        except SQLAlchemyError:
             session.rollback()
-            logger.error('Message status save failed', error=e)
+            logger.exception('Message status save failed')
```
necrophonic: Context?

secure_message/repository/saver.py
```diff
@@ -46,9 +47,9 @@ def save_msg_event(msg_id, event, session=db.session):
         try:
             session.add(event)
             session.commit()
-        except Exception as e:
+        except SQLAlchemyError:
             session.rollback()
-            logger.error('Message event save failed', error=e)
+            logger.exception('Message event save failed')
```
necrophonic: Context?

secure_message/repository/saver.py
```diff
@@ -60,7 +61,20 @@ def save_msg_audit(msg_id, msg_urn, session=db.session):
         try:
             session.add(db_audit)
             session.commit()
-        except Exception as e:
+        except SQLAlchemyError:
             session.rollback()
-            logger.error('Message audit save failed', error=e)
+            logger.exception('Message audit save failed')
```
necrophonic: Context?

secure_message/repository/saver.py
```diff
@@ -60,7 +61,20 @@ def save_msg_audit(msg_id, msg_urn, session=db.session):
...
+    def save_msg_actors(msg_id, from_actor, to_actor, sent_from_internal, session=db.session):
+        """Save actor information data to database"""
+        actor = Actors(msg_id, from_actor, to_actor, sent_from_internal)
+
+        try:
+            session.add(actor)
+            session.commit()
+        except SQLAlchemyError:
+            session.rollback()
+            logger.exception('Message actor save failed')
```
necrophonic: Context?

tests/app/test_saver.py
necrophonic: I mean the `pass` rather than the test itself :)

secure_message/repository/database.py
AndrewTorrance: The idea here is that the constant values are used for validation purposes and the +1 defines the column size so that  the column size can always cope with any value that passes validation even if we get a \0 added at some point .

secure_message/repository/database.py
AndrewTorrance: They return the data in a serialisable format . This pattern is used on all domain classes . AId rather leave it in until the team agrees on this one

secure_message/repository/database.py
necrophonic: Is this premature future proofing?

secure_message/repository/database.py
necrophonic: Probably should add an issue or something to deal with clarifying that going forward. Fair enough for consistency in the short term.

secure_message/repository/saver.py
AndrewTorrance: Its a valid point , I am not planning on changing it here , as I think there is some confusion as to what logging patterns we should be using . I am putting a separate ticket on our backlog to agree a format on our backlog .

secure_message/repository/saver.py
AndrewTorrance: See above

secure_message/repository/saver.py
AndrewTorrance: See above

secure_message/repository/saver.py
AndrewTorrance: see above

secure_message/repository/saver.py
AndrewTorrance: See above

tests/app/test_saver.py
AndrewTorrance: Fair cop , ( although we should also agree on this in the team ?) we should be consistent will change

AndrewTorrance: Changed as requested so will dismiss the review

tests/app/test_saver.py
AndrewTorrance: Just shoot me now. Changing ( with a red face)

secure_message/resources/messages.py
AndrewTorrance: No in the API . In Domain.py. In MessageSchema and DraftSchema ( different rules apply) all messages and drafts are validated close to the point of entry before passing closer to the DB

secure_message/repository/database.py
AndrewTorrance: Not future proofing .Its easier to add this defensively at the beginning knowing that validation of all external entities is done at sizes less than the column capacity than to never know if you get a column overflow if it was caused by data passed in or some muck up in code. By doing this I know that if I ever get an overflow error and I can proove that the validation is to less than that size , any overflow must be introduced by our code not the data passed in. To me it makes things an awful lot cleaner and easier to understand . I would argue its good practice

tests/app/test_saver.py
```diff
@@ -76,13 +77,52 @@ def test_save_msg_audit_does_not_raise_exception_on_successful_save(self):
...
                     Saver().save_msg_audit(message_audit['msg_id'], message_audit['msg_urn'], mock_session)
 
+    def test_save_actors_does_not_raise_exception_on_successful_save(self):
+        """Tests message actors is saved to database"""
+        message = {'msg_id': 'MsgId', 'msg_to': 'Tej', 'msg_from': 'some survey'}
+        with self.app.app_context():
+            with current_app.test_request_context():
+                Saver().save_message(SecureMessage(msg_id=message['msg_id']))
+                Saver().save_msg_actors(message['msg_id'], message['msg_to'], message['msg_from'], False)
+                self.assertTrue(True)
```
dcdarrell9: Are we asserting that True is True here?

tests/behavioural/features/steps/common.py
insacuri: Works for me, a comment explaining that would be useful but not essential

secure_message/resources/messages.py
```diff
@@ -89,14 +89,14 @@ def _message_save(message, is_draft, draft_id):
         """Saves the message to the database along with the subsequent status and audit"""
         save = Saver()
         save.save_message(message.data)
+        save.save_msg_actors(message.data.msg_id, message.data.msg_from, message.data.msg_to[0], g.user.is_internal)
```
JamesGardiner: Will `g` always have a `user` attribute? If not, what is the expected behaviour? As of Flask 0.10 `g` implements a `get` method that returns None by default.

AndrewTorrance: We always have it. When we read the jwt as part of the authentication we create a userobject and set  g.user  to be this user object. This is done prior to all calls , so we are safe then in reading it anywhere in the api

PR #137, opened 2018-02-05
created internal user mock

dcdarrell9:

[card](https://trello.com/c/UPRGrQhE)

Added an internal user service mock and toggle

**How to review**

Ensure service runs and there are no errors

### DIFF
--- a/secure_message/api_mocks/internal_user_service_mock.py
+++ b/secure_message/api_mocks/internal_user_service_mock.py
@@ -6,15 +6,15 @@ logger = wrap_logger(logging.getLogger(__name__))
 
 
 class InternalUserServiceMock:
-    @staticmethod
-    def get_user_details(uuid):
+    def __init__(self):
+        self.internal_user_dict = {"Someuuid": {"id": "Someuuid",
+                                                "firstName": "fred",
+                                                "lastName": "flinstone",
+                                                "emailAddress": "mock@email.com"}}
+
+    def get_user_details(self, uuid):
         """gets the user details from the internal user service"""
         logger.debug("getting mock user details for uaa")
         if uuid:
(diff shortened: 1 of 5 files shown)


secure_message/api_mocks/internal_user_service_mock.py
```diff
@@ -0,0 +1,20 @@
...
+    @staticmethod
+    def get_user_details(uuid):
+        """gets the user details from the internal user service"""
+        logger.debug("getting mock user details for uaa")
+        if uuid:
+            internal_user_dict = {"id": uuid,
+                                  "firstName": "fred",
+                                  "lastName": "flinstone",
+                                  "emailAddress": "mock@email.com"}
+            return internal_user_dict, 200
```
necrophonic: Do you need to assign `internal_user_dict` before returning it or could you return it directly without the assignment?

tests/behavioural/features/steps/common.py
```diff
@@ -44,6 +46,12 @@ def step_impl_use_mock_case_service(context):
     case_service.use_mock_service()
 
 
+@given("using mock internal user service")
+def step_impl_use_mock_internal_user_service(context):
+    """ use the default party service tests"""
```
necrophonic: 🎨 Minor cosmetic - consistency - should doc strings start with upper or lower case chars? Seems to have a mix here.

dcdarrell9: Should be done right 👍

secure_message/api_mocks/internal_user_service_mock.py
AndrewTorrance: This needs to accept a uuid and select the relevant dictionary entry . I think those tests we discussed yesterday will help ?

PR #140, opened 2018-02-08
Feature/support multiple internal users

AndrewTorrance:
This supports the changing of the database such that the existing tests all pass . It is a preparatory step prior to supporting multiple internal users. The reason for the PR is that we need a breaking change in order to support multiple users and hence we need to version the api .  
 

Added Actors column which supports identifying if an actor is internal or external. Added support for a message to a group by using the keyword 'GROUP' in the to field. ( could not use empty as that would fail validation in the case of a user not setting an actor)


secure_message/api_mocks/internal_user_service_mock.py
```diff
@@ -7,14 +7,50 @@
...
 
     def get_user_details(self, uuid):
         """gets the user details from the internal user service"""
         logger.debug("getting mock user details for uaa")
         if uuid:
-            return self.internal_user_dict[uuid], 200
-        return "error retrieving details"
+            found = self.internal_user_dict.get(uuid)
+            if found:
+                return found, 200
```
atrozzo: I would have done like : 
if  uuid in elf.internal_user_dict
  return true, 200

return False, 400 

just for consistency in the return type. and log with error/warning user not found.

secure_message/repository/retriever.py
```diff
@@ -16,16 +16,22 @@
...
     def retrieve_message_list(page, limit, user, ru_id=None, survey=None, cc=None, ce=None, label=None, descend=True):
         """returns list of messages from db"""
-        conditions = []
-        status_conditions = []
 
         if user.is_respondent:
-            status_conditions.append(Status.actor == str(user.user_uuid))
-        else:
-            status_conditions.append(Status.actor == constants.BRES_USER)
+            return Retriever._retrieve_message_list_respondent(page, limit, user=user, ru_id=ru_id, survey=survey,
```
atrozzo: Not sure about the python way to do things, but when there ar eso many parameters is not better to encapsulate them all in a Object ?

secure_message/repository/retriever.py
```diff
@@ -46,15 +52,73 @@ def retrieve_message_list(page, limit, user, ru_id=None, survey=None, cc=None, c
...
+                    .filter(SecureMessage.msg_id == t.c.msg_id) \
+                    .order_by(t.c.max_date.desc()).paginate(page, limit, False)
+
+            else:
+                result = SecureMessage.query \
+                    .filter(SecureMessage.msg_id == t.c.msg_id) \
+                    .order_by(t.c.max_date.asc()).paginate(page, limit, False)
+
+        except Exception as e:
+            logger.error('Error retrieving messages from database', error=e)
```
rhiGriff: Shouldn't this be a logger.exception?

secure_message/api_mocks/internal_user_service_mock.py
JamesGardiner: Can we try and use EAFP instead of LBYL? https://blogs.msdn.microsoft.com/pythonengineering/2016/06/29/idiomatic-python-eafp-versus-lbyl/

```Python
try:
    return self.internal_user_dict[uuid], 200
except KeyError:
    logger.exception(f'No internal user with uuid {uuid}'
    return "error retrieving details", 404
```

JamesGardiner: Also, don't inspect the uuid arg. If it's None then we'll still catch the exception, and if we set None as a key in a dictionary we have bigger problems.

secure_message/repository/database.py
```diff
@@ -100,11 +95,10 @@ def _populate_events(self, message):
             elif row.event == EventsApi.READ.value:
                 message['read_date'] = str(row.date_time)
 
-    def _populate_to_and_from(self, actor, message):
+    def _populate_to_and_from(self, user, message):
         for row in self.statuses:
-            if row.actor == actor:
+            if row.actor == user.user_uuid or (user.is_internal and (row.actor == constants.NON_SPECIFIC_INTERNAL_USER or row.actor == constants.BRES_USER)):
```
JamesGardiner: Is this here for the purposes of passing tests?

secure_message/repository/modifier.py
```diff
@@ -91,12 +91,14 @@ def del_draft(draft_id):
         del_draft_status = "DELETE FROM securemessage.status WHERE msg_id='{0}' AND label='{1}'".format(draft_id, Labels.DRAFT.value)
         del_draft_event = "DELETE FROM securemessage.events WHERE msg_id='{0}'".format(draft_id)
         del_draft_inbox_status = "DELETE FROM securemessage.status WHERE msg_id='{0}' AND label='{1}'".format(draft_id, Labels.DRAFT_INBOX.value)
+        del_actors = "DELETE FROM securemessage.actors where msg_id='{0}'".format(draft_id)
         del_draft_msg = "DELETE FROM securemessage.secure_message WHERE msg_id='{0}'".format(draft_id)
 
         try:
             db.get_engine(app=db.get_app()).execute(del_draft_status)
```
JamesGardiner: Isn't this where we should be using db.session?

secure_message/repository/modifier.py
```diff
@@ -91,12 +91,14 @@ def del_draft(draft_id):
...
             db.get_engine(app=db.get_app()).execute(del_draft_inbox_status)
             db.get_engine(app=db.get_app()).execute(del_draft_event)
+            db.get_engine(app=db.get_app()).execute(del_actors)
             db.get_engine(app=db.get_app()).execute(del_draft_msg)
 
         except Exception as e:
```
JamesGardiner: Probably worth changing this to logger.exception while we are at it.

secure_message/repository/retriever.py
JamesGardiner: This could be made cleaner by refactoring the `is not None` below to just do dict key access on kwargs passed here. If a KeyError is raised, you know the key isn't passed and therefore can do the `else` condition from below.

secure_message/repository/retriever.py
```diff
@@ -46,15 +52,73 @@ def retrieve_message_list(page, limit, user, ru_id=None, survey=None, cc=None, c
...
             t = db.session.query(SecureMessage.msg_id, func.max(Events.date_time)  # pylint:disable=no-member
-                                 .label('max_date'))\
+                                 .label('max_date')) \
                 .join(Events).join(Status).outerjoin(Actors) \
                 .filter(and_(*conditions)) \
                 .filter(and_(*status_conditions)) \
                 .filter(or_(Events.event == EventsApi.SENT.value, Events.event == EventsApi.DRAFT_SAVED.value)) \
                 .group_by(SecureMessage.msg_id).subquery('t')
 
             if descend:
```
JamesGardiner: If you follow my appraoch above set this to be a default kwarg.

secure_message/repository/retriever.py
```diff
@@ -46,15 +52,73 @@ def retrieve_message_list(page, limit, user, ru_id=None, survey=None, cc=None, c
...
+        if ce is not None:
+            conditions.append(SecureMessage.collection_exercise == str(ce))
+
+        try:
+            t = db.session.query(SecureMessage.msg_id, func.max(Events.date_time)  # pylint:disable=no-member
+                                 .label('max_date')) \
+                .join(Events).join(Status).outerjoin(Actors) \
+                .filter(and_(*conditions)) \
+                .filter(or_(*actor_conditions)) \
+                .filter(~Status.label.in_(status_reject_conditions)) \
```
JamesGardiner: Might be worth a comment here that SQLAlchemy overrides the unary bitwise operator

secure_message/repository/retriever.py
```diff
@@ -46,15 +52,73 @@ def retrieve_message_list(page, limit, user, ru_id=None, survey=None, cc=None, c
...
             if descend:
-                result = SecureMessage.query\
+                result = SecureMessage.query \
+                    .filter(SecureMessage.msg_id == t.c.msg_id) \
+                    .order_by(t.c.max_date.desc()).paginate(page, limit, False)
+
+            else:
+                result = SecureMessage.query \
+                    .filter(SecureMessage.msg_id == t.c.msg_id) \
+                    .order_by(t.c.max_date.asc()).paginate(page, limit, False)
```
JamesGardiner: Sqlalchemy allows query building, so the duplicated code here can by completely removed apart from the .asc or .desc.

Or just do

```Python
if descend:
    order = t.c.max_date.desc()
else:
    order = t.c.max_date.asc()
```

secure_message/validation/user.py
```diff
@@ -17,6 +17,11 @@ def is_respondent(self):
         return bool(self.role == 'respondent')
 
     @staticmethod
-    def is_valid_user(uuid):
+    def is_valid_internal_user(uuid):
+        _, status_code = internal_user_service.get_user_details(uuid)
+        return True if status_code == 200 else False
```
warrenbailey: can be simplified to

return status_code == 200

secure_message/repository/modifier.py
```diff
@@ -91,12 +91,14 @@ def del_draft(draft_id):
...
         del_draft_event = "DELETE FROM securemessage.events WHERE msg_id='{0}'".format(draft_id)
         del_draft_inbox_status = "DELETE FROM securemessage.status WHERE msg_id='{0}' AND label='{1}'".format(draft_id, Labels.DRAFT_INBOX.value)
+        del_actors = "DELETE FROM securemessage.actors where msg_id='{0}'".format(draft_id)
         del_draft_msg = "DELETE FROM securemessage.secure_message WHERE msg_id='{0}'".format(draft_id)
 
         try:
             db.get_engine(app=db.get_app()).execute(del_draft_status)
             db.get_engine(app=db.get_app()).execute(del_draft_inbox_status)
             db.get_engine(app=db.get_app()).execute(del_draft_event)
+            db.get_engine(app=db.get_app()).execute(del_actors)
```
warrenbailey: could we just call get_engine once?

secure_message/validation/user.py
```diff
@@ -17,6 +17,11 @@ def is_respondent(self):
...
+
+    @staticmethod
+    def is_valid_respondent(uuid):
         _, status_code = party.get_user_details(uuid)
-        return status_code == 200
+        return True if status_code == 200 else False
```
warrenbailey: As above

secure_message/api_mocks/internal_user_service_mock.py
AndrewTorrance: Normally I would support EAFP instead of LBYL . However in this case the existing case and party mocks are written so as to mimic the http response which would not raise an exception to the code  but return an error . In this specific of the internal_user_service_mock it is something of a guess to assume it will be a http response . If it isnt and we do get exceptions I will change the mock implementation to suit. The uuid in elf.internal_user_dict needs to look in the values()

secure_message/repository/retriever.py
AndrewTorrance: I agree with possibly moving the parameters to kwargs . However this how the existing code handles it. So whilst I agree its best to follow the boy scout principle , in this case I will raise a ticket to loop back depending on time.

secure_message/repository/retriever.py
AndrewTorrance: Good catch , I copied existing code , changing

secure_message/repository/database.py
AndrewTorrance: No , its needed so that the status's are retrieved correctly in the case that a respondent sends a message to a group . If the user is internal we want to see the status's even if the message was sent to a group .

secure_message/repository/modifier.py
AndrewTorrance: Possibly , I must admit I followed what was there. Not sure of the implications of changing to that . The draft is normally deleted in the context of sending the same message as a draft .Other functions in the same class do use the session so Its possible that the two might interfere with each other , but unless I did an investigation I could not be sure .

secure_message/repository/modifier.py
AndrewTorrance: done

secure_message/repository/retriever.py
AndrewTorrance: I have raised a ticket to look at cleaning this up as mentioned above . I agree it can be improved

secure_message/repository/retriever.py
AndrewTorrance: Done

secure_message/repository/retriever.py
AndrewTorrance: Good catch , done

secure_message/validation/user.py
AndrewTorrance: doh , thanks , done

secure_message/repository/modifier.py
AndrewTorrance: I must admit I am not sure , I followed on from what was there before . I can look at this

secure_message/validation/user.py
AndrewTorrance: done

secure_message/api_mocks/internal_user_service_mock.py
JamesGardiner: My suggestion doesn't change the return values, only the way in which the decision for what values to return is made

secure_message/repository/modifier.py
JamesGardiner: Session should be used. It encapsulates db.get_engine. See https://github.com/mitsuhiko/flask-sqlalchemy/issues/107 for it's implementation.

secure_message/repository/modifier.py
warrenbailey: Think James comment above supersedes this now

secure_message/repository/modifier.py
```diff
@@ -91,16 +91,18 @@ def del_draft(draft_id):
...
             db.get_engine(app=db.get_app()).execute(del_draft_msg)
 
         except Exception as e:
-            logger.error('Error deleting draft from database', msg_id=draft_id, error=e)
+            logger.exception('Error deleting draft from database', msg_id=draft_id, error=e)
```
warrenbailey: logger.exception looks up the stack trace to find the exception so error=e is not needed

secure_message/v1/application.py
```diff
@@ -0,0 +1,30 @@
...
+from secure_message.resources.labels import Labels
+from secure_message.resources.messages import MessageList, MessageSend, MessageById, MessageModifyById
+from secure_message.resources.drafts import DraftSave, DraftById, DraftModifyById, DraftList
+from secure_message.resources.threads import ThreadById, ThreadList
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def set_v1_resources(api):
+    api.add_resource(Health, '/health')
```
JamesGardiner: Are these going to be updated following the tech session review yesterday?

secure_message/repository/retriever.py
warrenbailey: Why not use the named tuple created in the calling function instead? It seems to create the Named Tuple when the request arrives but then unpack it to seperate variables in order to call these function when actually everything you need (apart from the user) is in the tuple

secure_message/repository/retriever.py
```diff
@@ -46,38 +52,95 @@ def retrieve_message_list(page, limit, user, ru_id=None, survey=None, cc=None, c
...
+                .filter(SecureMessage.msg_id == t.c.msg_id) \
+                .order_by(order).paginate(page, limit, False)
 
         except Exception as e:
             logger.error('Error retrieving messages from database', error=e)
             raise InternalServerError(description="Error retrieving messages from database")
 
         return True, result
 
+    @staticmethod
```
warrenbailey: I'm really confused to why these methods return True, result?

The calling function does an if on the status which is always true?

secure_message/repository/retriever.py
```diff
@@ -46,38 +52,95 @@ def retrieve_message_list(page, limit, user, ru_id=None, survey=None, cc=None, c
...
+            result = SecureMessage.query \
+                .filter(SecureMessage.msg_id == t.c.msg_id) \
+                .order_by(order).paginate(page, limit, False)
+
+        except Exception as e:
+            logger.exception('Error retrieving messages from database', error=e)
+            raise InternalServerError(description="Error retrieving messages from database")
+
+        return True, result
+
```
warrenbailey: I'm really confused to why these methods return True, result?

The calling functions do an if on the status but its always true so appears to be pointless, can we just remove it and return the result?

secure_message/repository/retriever.py
```diff
@@ -46,38 +52,95 @@ def retrieve_message_list(page, limit, user, ru_id=None, survey=None, cc=None, c
...
         """Count users unread messages"""
         status_conditions = []
         status_conditions.append(Status.actor == str(user.user_uuid))
         try:
-            result = SecureMessage.query.join(Status).filter(and_(*status_conditions)).filter(Status.label == 'UNREAD').count()
+            result = SecureMessage.query.join(Status).filter(and_(*status_conditions)).filter(
+                Status.label == 'UNREAD').count()
         except Exception as e:
-            logger.error('Error retrieving count of unread messages from database', error=e)
+            logger.exception('Error retrieving count of unread messages from database', error=e)
```
warrenbailey: as above no need to log error=e

secure_message/repository/retriever.py
```diff
@@ -46,38 +52,95 @@ def retrieve_message_list(page, limit, user, ru_id=None, survey=None, cc=None, c
...
+        """returns list of messages from db"""
+        conditions = []
+        status_reject_conditions = []
+        valid_statuses = []
+        actor_conditions = []
+
+        if label is not None:
+            valid_statuses.append(label)
+            if label in [Labels.INBOX.value, Labels.ARCHIVE.value, Labels.UNREAD.value]:
+                actor_conditions.append(Actors.sent_from_internal == False)  # NOQA pylint:disable=singleton-comparison
```
warrenbailey: This to me sums up why the DB structure isn't correct because it shouldn't be this complicated to determine if the internal user has sent or received the message

secure_message/v1/application.py
warrenbailey: These are the v1 resources aren't they?

secure_message/api_mocks/internal_user_service_mock.py
AndrewTorrance: ok , misunderstood , changed

secure_message/repository/modifier.py
AndrewTorrance: done

secure_message/v1/application.py
AndrewTorrance: If you mean the health endpoint , I am not aware of any request to change that

secure_message/repository/retriever.py
AndrewTorrance: Good catch . Me too , changed.

secure_message/repository/retriever.py
AndrewTorrance: As above , done

secure_message/repository/retriever.py
AndrewTorrance: done

secure_message/repository/retriever.py
AndrewTorrance: Thats not what its doing . Its adding predicates specifically only when someone has asked for only messages with specific labels to be returned . The simple sent_from_internal flag tells us if it was sent by an internal person

secure_message/repository/retriever.py
AndrewTorrance: I am putting this on a tech debt ticket that I plan on coming back to asap

secure_message/repository/modifier.py
AndrewTorrance: Added as a tech debt issue

secure_message/repository/modifier.py
AndrewTorrance: tech debt ticket added

secure_message/repository/retriever.py
AndrewTorrance: Actually tried to change as I agree it looks wrong . Get lots of failing tests . Raised a tech debt ticket . Plan on looping back to this

secure_message/repository/retriever.py
warrenbailey: Seems I misunderstood it then

warrenbailey:
Approved as tech debt tickets have been raised to fix the other comments and the understanding they'll be address asap in a separate PR.

PR #143, opened 2018-02-16
Database resilience

benjefferies:

Link to Trello card

Adding retry mechanism for when service is up before database. It means that the service will try every 2 seconds for 30 seconds to connect

**How to review**

Run service without database, start database after a few seconds and it should connect

### DIFF
--- a/secure_message/application.py
+++ b/secure_message/application.py
@@ -38,15 +38,7 @@ def create_app(config=None):
     api = Api(app)
     CORS(app)
 
-    database.db.init_app(app)
-
-    logger.info('Starting Secure Message Service...', config=app_config)
-
-    with app.app_context():
-        event.listen(database.db.metadata, 'before_create', DDL(
-            "CREATE SCHEMA IF NOT EXISTS securemessage"))
-        database.db.create_all()
-        database.db.session.commit()  # NOQA pylint:disable=no-member
+    create_db(app, app_config)
 
     api.add_resource(Health, '/health')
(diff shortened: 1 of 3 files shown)


secure_message/application.py
```diff
@@ -83,6 +77,22 @@ def handle_exception(error):  # NOQA pylint:disable=unused-variable
...
 
+def retry_if_database_error(exception):
+    logger.error('Database error has occurred', error=exception)
+    return isinstance(exception, DatabaseError) and not isinstance(exception, ProgrammingError)
+
+
+@retry(retry_on_exception=retry_if_database_error, wait_fixed=2000, stop_max_delay=30000, wrap_exception=True)
+def create_db(app, app_config):
+    database.db.init_app(app)
+    logger.info('Starting Secure Message Service...', config=app_config)
```
jcox-dev: This log line can probably live outside of the create_db function.

benjefferies: 👍 Done

JamesGardiner:
https://pypi.python.org/pypi/retrying

benjefferies:
@JamesGardiner Fair point. There is this https://github.com/rholder/retrying/pull/72 and https://travis-ci.org/rholder/retrying/builds/287235053?utm_source=github_status&utm_medium=notification. I have left a comment on the PR to see if it could be moved forward but whether we want to use it without 3.6 in the travis test matrix on master yet is the question?

JamesGardiner:
@benjefferies considering it has had a green build will approve for now.

PR #144, opened 2018-02-19
Feature/version two endpoints removing actor

AndrewTorrance:
Adding the Version 2 end points . 

 Changing the retriever functions to not return a tuple, using the mock internal user service . Removing the actors table, adding from internal to secure message so that we know if the sender is internal.

new endpoint names defined in application.py
internal user lookup added to message/draft/thread retrieval
changed secure message table , dropped actors table, dropped internal sent audit table 
changed labels endpoint to message counter for clarity , retained original in V1.
changed retriever to return just data , not a confusing tuple
Added behavioural tests for new endpoints


secure_message/common/utilities.py
```diff
@@ -143,6 +143,43 @@ def get_details_by_uuids(uuids):
...
+        to_user_details_finder = internal_user_service.get_user_details
+
+    for uuid in message['msg_to']:
+
+        detail, status_code = to_user_details_finder(uuid)
+
+        if status_code == 200:
+            user_details.append(detail)
+        else:
+            logger.info('No details found for user', uuid=uuid)
```
gemmai95: I think this should be party_id rather than uuid as that doesn't really mean anything.

secure_message/resources/messages.py
```diff
@@ -35,16 +35,15 @@ def get():
...
-                                                               cc=message_args.cc, label=message_args.label,
-                                                               descend=message_args.desc, ce=message_args.ce)
-
-        if status:
-            resp = paginated_list_to_json(result, message_args.page, message_args.limit, request.host_url,
-                                          g.user, message_args.string_query_args, MESSAGE_LIST_ENDPOINT)
-            resp.status_code = 200
-            return resp
+        result = message_service.retrieve_message_list(message_args.page, message_args.limit, g.user,
+                                                       ru_id=message_args.ru_id, survey=message_args.survey,
```
gemmai95: In the rest of the services this is referred to as ru_ref or business_id.

secure_message/v2/resources/messages.py
```diff
@@ -0,0 +1,53 @@
+from flask import request, jsonify, g
+from flask_restful import Resource
+from werkzeug.exceptions import BadRequest
+
+from secure_message.common.eventsapi import EventsApi
+from secure_message.common.labels import Labels
+from secure_message.repository.saver import Saver
+from secure_message.resources.messages import MessageSend
+from secure_message.v2.validation.domain import MessageSchemaV2, logger
+from secure_message.v2.repository.retriever import RetrieverV2
```
gemmai95: Alphabetical imports

tests/app/test_app.py
```diff
@@ -392,6 +377,7 @@ def test_case_service_called_on_sent_if_NotifyCaseService_is_set(self, case):
 
         self.app.config["NOTIFY_VIA_GOV_NOTIFY"] = '0'
         self.app.config['NOTIFY_CASE_SERVICE'] = '1'
+        self.app.config['NOTIFY_VIA_GOV_NOTIFY'] = '0'
```
gemmai95: This is the same as one of the lines above.

secure_message/common/utilities.py
AndrewTorrance: God catch , this code should have been deleted , its been superseded . Deleting .

secure_message/resources/messages.py
AndrewTorrance: It needs a wider discussion as to ru-id/ru ref and business ref which is outside of the scope of this pr and one for the tech leads

secure_message/v2/resources/messages.py
AndrewTorrance: Done

tests/app/test_app.py
AndrewTorrance: Yep , came in during a merge

tests/behavioural/features/v2/draft_get.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Get draft by id V2
+
+  Background: Reset database
+    Given prepare for tests using 'mock' services
+
+  Scenario Outline: A Respondent saves a draft and reads it
+      And sending from respondent to internal <user>
```
dcdarrell9: In the V2 feature files I think a lot of these should be split out, I don't think having 5+ 'and' is helpful. If we're using that many then they are really different scenarios

secure_message/application.py
```diff
@@ -42,7 +47,28 @@ def create_app(config=None):
...
+
+    api.add_resource(DraftSave, '/draft/save', '/v2/drafts')
+    api.add_resource(DraftModifyById, '/draft/<draft_id>/modify', '/v2/drafts/<draft_id>')
+    api.add_resource(DraftById, '/draft/<draft_id>', '/v2/drafts/<draft_id>')
+    api.add_resource(DraftList, '/drafts', '/v2/drafts')
+
+    api.add_resource(ThreadList, '/threads')
+    api.add_resource(ThreadById, '/thread/<thread_id>', '/v2/threads/<thread_id>')
+
+    # V2
```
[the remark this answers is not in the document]
AndrewTorrance: Changing comments

secure_message/application.py
```diff
@@ -8,11 +8,16 @@
...
+from secure_message.exception.exceptions import MissingEnvironmentVariable
 from secure_message.logger_config import logger_initial_config
-from secure_message.v1.application import set_v1_resources
+from secure_message.repository import database
+from secure_message.resources.drafts import DraftSave, DraftById, DraftModifyById, DraftList
+from secure_message.resources.health import Health, DatabaseHealth, HealthDetails
+from secure_message.resources.info import Info
+from secure_message.resources.messages import MessageList, MessageSend, MessageById, MessageModifyById, MessageCounter
+from secure_message.v2.resources.messages import MessageSendV2, MessageCounterV2
+from secure_message.resources.threads import ThreadById, ThreadList
```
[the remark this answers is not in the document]
AndrewTorrance: Done , not pushed

tests/behavioural/features/v2/draft_get.feature
AndrewTorrance: Agree with complexity is potentially an issue , but not sure how to make them easier given the complexity around threads

secure_message/common/utilities.py
JamesGardiner: Logging should be done as close to the code as possible. In this case, anything other than a 200 should be handled in to_user_details_finder. If you deal with HTTPErrors there, a lot of this function could be done as:

```python
[to_user_details_finder(uuid)[0] for uuid in message['msg_to']]
```

secure_message/common/utilities.py
```diff
@@ -143,22 +143,39 @@ def get_details_by_uuids(uuids):
...
-    uuid_list = []
 
-    for message in messages:
-        uuid_list.extend([uuid for uuid in message['msg_to'] if uuid not in uuid_list])
-        if message['msg_from'] not in uuid_list:
-            uuid_list.append(message['msg_from'])
+def _get_from_details(message):
+    """looks up the details for the from users"""
+    if message['from_internal']:
+        from_details, _ = internal_user_service.get_user_details(message['msg_from'])
```
JamesGardiner: Get user details shouldn't be returning a tuple - it's not used anywhere and makes the code overly verbose. If you need to handle HTTPErrors deal with them in the calling code or let the exception move up the stack

secure_message/repository/retriever.py
```diff
@@ -280,6 +280,4 @@ def check_msg_id_is_a_draft(draft_id, user):
             logger.error('Error retrieving message from database', error=e)
             raise InternalServerError(description="Error retrieving message from database")
 
-        if result is None:
-            return False, result
-        return True, result.serialize(user)
+        return result.serialize(user) if result else None
```
JamesGardiner: Why are you checking if result evaluates to True? If you move this in to the try block as:

```Python
return result.serialize(user)
```

it will still work, as result will always be an SqlAlchemy object or an exception will be thrown.

secure_message/repository/saver.py
```diff
@@ -3,7 +3,7 @@
 from sqlalchemy.exc import SQLAlchemyError
 from structlog import wrap_logger
 from secure_message.exception.exceptions import MessageSaveException
-from secure_message.repository.database import Actors, db, SecureMessage, Status, Events, InternalSentAudit
+from secure_message.repository.database import db, SecureMessage, Status, Events
```
JamesGardiner: Events should be between db and SecureMessage

secure_message/resources/messages.py
```diff
@@ -76,20 +77,34 @@ def post(self):
...
+
         if message.errors == {}:
-            return self._message_save(message, is_draft, draft_id)
+            self._message_save(message, is_draft, draft_id)
+            if is_draft:
+                Modifier().del_draft(draft_id)
+            # listener errors are logged but still a 201 reported
+            MessageSend._alert_listeners(message.data)
+            resp = jsonify({'status': '201', 'msg_id': message.data.msg_id, 'thread_id': message.data.thread_id})
+            resp.status_code = 201
```
JamesGardiner: Status is being set twice.

secure_message/resources/messages.py
```diff
@@ -247,3 +252,21 @@ def _validate_request(request_data):
...
 
         return action, label
+
+
+class MessageCounter(Resource):
+    @staticmethod
+    def get():
+        """Get count of unread messages"""
+
+        if request.args.get('name'):
```
JamesGardiner:
```python
try:
    if request.args['name'].lower() == 'unread':
        return jsonify(name=name, total=message_service.unread_message_count(g.user))
    else:
         logger.debug('Invalid label name', name=name, request=request.url)
         raise BadRequest(description="Invalid label"
except KeyError:
     logger.debug('No Name parameter specified in URL', request=request.url)
     raise BadRequest(description='No Label Name Parameter specified.')
```

JamesGardiner:
Reviewed up to messages.py - I'll pick up more tomorrow

secure_message/v2/repository/retriever.py
```diff
@@ -0,0 +1,51 @@
...
+                filter(or_(*status_conditions)). \
+                filter(and_(*survey_conditions)). \
+                filter(Status.label == 'UNREAD').count()
+        except Exception as e:
+            logger.error('Error retrieving count of unread messages from database', error=e)
+            raise InternalServerError(description="Error retrieving count of unread messages from database")
+        return result
+
+    @staticmethod
+    def _get_conditions_internal_user(survey, user):
```
insacuri: What do you mean by conditions?  I might've been told and forgotten but I think what these 2 functions do could maybe warrant a comment for clarity

secure_message/common/utilities.py
```diff
@@ -143,22 +143,39 @@ def get_details_by_uuids(uuids):
 
 def add_to_and_from_details(messages):
     """Adds user details for sender and recipient"""
+    for message in messages:
```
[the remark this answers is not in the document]
AndrewTorrance: Good call , thanks

secure_message/common/utilities.py
AndrewTorrance: In the first comment its logging twice , the lower and where you see it . Removing the higher one along with the tuple return

gemmai95:
Also have the current Secure Messaging stories in review been tested with this functionality?

secure_message/application.py
```diff
@@ -37,7 +42,28 @@ def create_app(config=None):
...
+    api.add_resource(DraftSave, '/draft/save', '/v2/drafts')
+    api.add_resource(DraftModifyById, '/draft/<draft_id>/modify', '/v2/drafts/<draft_id>')
+    api.add_resource(DraftById, '/draft/<draft_id>', '/v2/drafts/<draft_id>')
+    api.add_resource(DraftList, '/drafts', '/v2/drafts')
+
+    api.add_resource(ThreadList, '/threads')
+    api.add_resource(ThreadById, '/thread/<thread_id>', '/v2/threads/<thread_id>')
+
+    # V2
+    api.add_resource(MessageSendV2, '/v2/messages')
```
gemmai95: This is the same as the above for message list

tests/app/test_draft.py
```diff
@@ -249,16 +249,16 @@ def test_draft_modified_since_last_read_true(self):
 
         with self.app.app_context():
             with current_app.test_request_context():
-                is_valid_draft = Retriever().check_msg_id_is_a_draft(msg_id, self.user_respondent)
-        self.assertTrue(is_valid_draft[0])
+                draft = Retriever().get_draft(msg_id, self.user_respondent)
```
JamesGardiner: Are these static methods on the Retriever class? If so, we can get rid of the `()`.

tests/behavioural/features/v2/draft_get.feature
JamesGardiner: I agree with @dcdarrell9 on this one - is there a way to specify tests against v1/v2 without the need to have so many `and` clauses?

AndrewTorrance: The two things are separate . Some of the tests can have reduced numbers of steps by combining some . But in the case of a conversation there are many things to do. Set up the user , set up the two and from , send a message , switch users , read a message , respond , possibly save a draft . I think some can be cleaned up on a case by case basis but there is no one size fits all .

secure_message/repository/retriever.py
AndrewTorrance: Done

secure_message/common/utilities.py
AndrewTorrance: Also changing party and user lookups to return None in the case of no data and log errors closer to the error detection.

secure_message/repository/saver.py
AndrewTorrance: Done

AndrewTorrance: BTW do you have a tool for detecting these  alphabetical issues ?

secure_message/resources/messages.py
AndrewTorrance: No, the jsonify sets status_code to 200 , we want a 201 returned to the user to indicate that its been added

JamesGardiner: Jsonify just wraps the JSON library with some helper methods that output a flask.Response object. I can see in line 89 that `"status": "201"` and 
 on line 90 `"status_code": "201"`. If you need to set both fields to the same value then there is something wrong, either here or elsewhere in the service.

secure_message/repository/saver.py
JamesGardiner: Mk1 human eyeball 👁 👍

secure_message/repository/retriever.py
AndrewTorrance: Actually , not quite . The serializer blows when zero records are returned, and i Need to distinguish if a message id was a draft in the case that someone attempts a put on a message id 
that is not a draft

secure_message/repository/database.py
```diff
@@ -71,21 +74,16 @@ def serialize(self, user, body_summary=False):
                    'ru_id': self.ru_id,
                    'survey': self.survey,
                    'collection_exercise': self.collection_exercise,
```
gemmai95: Is this an optional parameter? As it currently isn't used or a requirement

secure_message/v2/repository/retriever.py
AndrewTorrance: Adding comments . These functions set the query conditions/predicates used in the query to get a count of messages when the actor is internal . There is a corresponding one for external . 
Having said that I think I can see a bug .

secure_message/common/utilities.py
```diff
@@ -115,50 +115,47 @@ def get_business_details_by_ru(rus):
...
     for ru in rus:
 
-        detail, status_code = party.get_business_details(ru)
+        detail = party.get_business_details(ru)
 
-        if status_code == 200:
+        if detail is not None:
             details.append(detail)
         else:
             logger.info('No details found for RU ID', ru=ru)
```
gemmai95: This should be ru_ref or ru_id rather than just ru for the logging.

tests/app/test_app.py
```diff
@@ -464,7 +449,7 @@ def test_if_user_unknown_in_party_service_no_email_sent(self, mock_alerter, mock
                                                                        "lastName": "",
                                                                        "telephone": "+443069990888",
                                                                        "status": "ACTIVE",
-                                                                       "sampleUnitType": "BI"}, 200))
+                                                                       "sampleUnitType": "BI"}))
```
gemmai95: Why is sample unit type necessary, as secure message will never deal with the B case only BI?

PR #148, opened 2018-02-23
Improve authenticator by catching exceptions thrown.

insacuri:
Before this when the authenticator failed there would be no logging to show what was going wrong
because while exceptions were being thrown, nothing was catching them.

Link to Trello card

**How to review**


AndrewTorrance:
No need to catch BadRequest it propagates as a 400

PR #150, opened 2018-02-26
Modified case service so if no case id , then logs an error and does …

AndrewTorrance:
The case id should be passed by the ui , but if it is not then we should create useful error messages . 
Currently if the ui does not set a collection case then the error message is misleading.

It should log a useful error and return 

https://trello.com/c/j7T1bq0I

Case service checks for a case_id before sending a case event , logs an error if none present.


secure_message/resources/messages.py
```diff
@@ -137,24 +137,42 @@ def _try_send_alert_email(message):
...
-                    first_name = party_data['firstName'] if 'firstName' in party_data else ''
-                    last_name = party_data['lastName'] if 'lastName' in party_data else ''
-                    case_user = '{} {}'.format(first_name, last_name).strip()
-                    if not case_user:
-                        case_user = 'Unknown user'
-                        logger.info('no user names in party data for id  Unknown user used in case ',
-                                    party_id=party_data['id'])
-                # else not testable , as it fails on validation
-            case_service.store_case_event(message.collection_case, case_user)
+                logger.info(f'unable to resolve details, for case service, for {g.user.user_uuid} role: {g.user.role}')
```
atrozzo: Is it the case is a warning level ?

secure_message/resources/messages.py
```diff
@@ -137,24 +137,42 @@ def _try_send_alert_email(message):
...
+
         else:
             logger.info('Case service notifications switched off, hence not sent', msg_id=message.msg_id)
 
+    @staticmethod
+    def _resolve_user_details_for_case_service(user, message):
+        case_user = ''
+        if message.msg_from == constants.BRES_USER:
+            case_user = constants.BRES_USER
+        else:
```
atrozzo: The else condition could a private method _get_user_data or aggregate in _get_case_user. I assume user_data is goign to be always with some data inside.

secure_message/resources/messages.py
```diff
@@ -137,24 +137,42 @@ def _try_send_alert_email(message):
...
+                user_data = party.get_user_details(message.msg_from)  # NOQA TODO avoid 2 lookups(see validate)
+                service = 'party'
+
+            if user_data:
+                first_name = user_data.get('firstName', '')
+                last_name = user_data.get('lastName', '')
+                case_user = '{} {}'.format(first_name, last_name).strip()
+                if not case_user:
+                    case_user = 'Unknown user'
+                    logger.info(f'no user names in {service} service for id {user.user_uuid} Unknown user used in case',
```
atrozzo: As well is this a info or a warning level ?

secure_message/resources/messages.py
AndrewTorrance: Changed

secure_message/resources/messages.py
AndrewTorrance: comment out of date

secure_message/resources/messages.py
AndrewTorrance: Good call, changed

secure_message/resources/messages.py
```diff
@@ -137,24 +137,42 @@ def _try_send_alert_email(message):
...
+            case_user = constants.BRES_USER
+        else:
+            if g.user.is_internal:
+                user_data = internal_user_service.get_user_details(user.user_uuid)
+                service = 'user'
+            else:
+                user_data = party.get_user_details(message.msg_from)  # NOQA TODO avoid 2 lookups(see validate)
+                service = 'party'
+
+            if user_data:
```
atrozzo: No sure you need this if here. user_data is going to have always something in there

secure_message/services/case_service.py
```diff
@@ -11,16 +11,21 @@ class CaseService:
...
+                         reason=case_service_data.reason, text=case_service_data.text)
+            case_service_dict = json.loads(case_service_data.text)
+            if 'error' in case_service_dict:
+                case_service_dict = case_service_dict['error']
 
-        return case_service_dict, case_service_data.status_code
+            return case_service_dict, case_service_data.status_code
+        else:
+            err = f"No case id for case involving user {user_name}, case event not called"
+            logger.error(err)
```
dcdarrell9: 🎨 Could have this in one line

secure_message/resources/messages.py
```diff
@@ -137,24 +137,42 @@ def _try_send_alert_email(message):
...
+                first_name = user_data.get('firstName', '')
+                last_name = user_data.get('lastName', '')
+                case_user = '{} {}'.format(first_name, last_name).strip()
+                if not case_user:
```
atrozzo: Not sure if case_user is going to be ever empty here. 

I would aggregate all these functionality in a private method _get_user_data. The Unknow user should be the default one declared at the top instad of the empty String .

secure_message/resources/messages.py
```diff
@@ -137,24 +137,42 @@ def _try_send_alert_email(message):
...
+                user_data = internal_user_service.get_user_details(user.user_uuid)
+                service = 'user'
+            else:
+                user_data = party.get_user_details(message.msg_from)  # NOQA TODO avoid 2 lookups(see validate)
+                service = 'party'
+
+            if user_data:
+                first_name = user_data.get('firstName', '')
+                last_name = user_data.get('lastName', '')
+                case_user = '{} {}'.format(first_name, last_name).strip()
```
insacuri: 🎨 Use an f string here, looks cleaner

secure_message/services/case_service.py
```diff
@@ -11,16 +11,21 @@ class CaseService:
     def store_case_event(case_id, user_name):
         """posts the data to the case service"""
 
-        json_data = {"description": "New Secure Message",
-                     "category": "SECURE_MESSAGE_SENT",
-                     "createdBy": user_name}
+        if case_id:
+            json_data = {"description": "New Secure Message",
+                         "category": "SECURE_MESSAGE_SENT",
+                         "createdBy": user_name}
```
insacuri: What happens if the user_name is somehow empty?  Should there be some checking for this if it's important?

tests/app/test_message.py
```diff
@@ -17,6 +17,9 @@ class MessageTestCase(unittest.TestCase):
     def setUp(self):
         """setup test environment"""
         self.now = datetime.now(timezone.utc)
+        internal_user_service.use_mock_service()
```
dcdarrell9: Is this being used anywhere?

secure_message/services/case_service.py
AdamHawtin: This is being handled where this function is called in the _inform_case_service which only calls this if it has a non empty name.

secure_message/resources/messages.py
```diff
@@ -137,24 +137,42 @@ def _try_send_alert_email(message):
...
-                # else not testable , as it fails on validation
-            case_service.store_case_event(message.collection_case, case_user)
+                logger.warning(f'unable to resolve details, for case service, for {g.user.user_uuid} role: {g.user.role}')
+
         else:
             logger.info('Case service notifications switched off, hence not sent', msg_id=message.msg_id)
 
+    @staticmethod
+    def _resolve_user_details_for_case_service(user, message):
+        case_user = ''
```
JamesGardiner: This is overly verbose and not pythonic

secure_message/services/case_service.py
AndrewTorrance: no because it logs the error then returns it

secure_message/resources/messages.py
AndrewTorrance: user data can be none if it is not known in party

secure_message/resources/messages.py
AndrewTorrance: Good idea , done

secure_message/resources/messages.py
AndrewTorrance: done

secure_message/resources/messages.py
AndrewTorrance: fair cop, tidying it up

secure_message/validation/domain.py
```diff
@@ -169,10 +169,10 @@ def check_variables_set_and_not_set(self, data):   # NOQA pylint:disable=no-self
         """Check sent and read date not set and that from and survey are set"""
         if 'msg_from' not in data or not data['msg_from']:
```
JamesGardiner: This should be:

```Python
try:
    [data[k] for k in ['msg_from', 'survey']]
except KeyError as e:
	logger.exception(f"{e.args[0]} field missing')
else:
    return data
```

tests/app/test_app.py
```diff
@@ -451,9 +452,9 @@ def test_if_user_unknown_in_party_service_no_email_sent(self, mock_alerter, mock
                                                                        "status": "ACTIVE",
                                                                        "sampleUnitType": "BI"}))
     @patch.object(CaseServiceMock, 'store_case_event')
-    @patch.object(message_logger, 'info')
-    def test_if_user_has_no_first_name_or_last_name_then_unknown_user_passed_to_case_service(self, mock_logger,
-                                                                                             mock_case, mock_party):
+    @patch.object(message_logger, 'warning')
```
JamesGardiner: Why patch the logger when there is a perfectly good assertLogs method in unittest? https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs

tests/app/test_app.py
```diff
@@ -479,10 +480,40 @@ def test_if_user_has_no_first_name_or_last_name_then_unknown_user_passed_to_case
...
         mock_case.assert_called_with('ACollectionCase', 'Unknown user')
-        mock_logger.assert_called_with('no user names in party data for id  Unknown user used in case ',
-                                       party_id='f62dfda8-73b0-4e0e-97cf-1b06327a6712')
 
+    @patch.object(InternalUserServiceMock, 'get_user_details', return_value=({"id": "f62dfda8-73b0-4e0e-97cf-1b06327a6712",
+                                                                              "emailAddress": "   ",
+                                                                              "lastName": "",
+                                                                              "telephone": "+443069990888"}))
+    @patch.object(CaseServiceMock, 'store_case_event')
+    @patch.object(message_logger, 'warning')
```
JamesGardiner: Is this patch even needed?

secure_message/services/case_service.py
```diff
@@ -11,16 +11,21 @@ class CaseService:
     def store_case_event(case_id, user_name):
         """posts the data to the case service"""
 
-        json_data = {"description": "New Secure Message",
-                     "category": "SECURE_MESSAGE_SENT",
-                     "createdBy": user_name}
+        if case_id:
```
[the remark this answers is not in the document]
AndrewTorrance: Yes , if the ui cant resolve it , which is the issue that Adam and Angelo currently have. The call to case was having a double slash when no id and the error reported was incorrect verb, which is the reason we are doing the change .

secure_message/resources/messages.py
```diff
@@ -137,24 +137,30 @@ def _try_send_alert_email(message):
...
+        if message.msg_from == constants.BRES_USER:
+            user_name = constants.BRES_USER
+        else:
+            user_data = internal_user_service.get_user_details(message.msg_from) if user.is_internal else party.get_user_details(message.msg_from)
+
+            if user_data:
+                first_name = user_data.get('firstName', '')
+                last_name = user_data.get('lastName', '')
+                full_name = f'{first_name} {last_name}'.strip()
+                if full_name:
```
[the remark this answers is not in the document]
AndrewTorrance: What if the full name is a space ? I have no control over how they are stored in party or the uaa so there may be spaces

secure_message/validation/domain.py
AndrewTorrance: Not quite, this this does not detect if a msg_from or survey field is present but with no value

tests/app/test_app.py
AndrewTorrance: nope , not needed , removed . I cant even offer an explanation as to why on earth that was there

secure_message/resources/messages.py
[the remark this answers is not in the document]
JamesGardiner: A space will still not evaluate to None.

secure_message/services/case_service.py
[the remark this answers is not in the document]
JamesGardiner: In that case, shouldn't we be checking that there is a case_id before calling this function?

secure_message/services/case_service.py
JamesGardiner: Is it also checking for a not None `case_id`?

secure_message/validation/domain.py
JamesGardiner: When would a msg_from ever be empty? A message is always sent by a user, isn't it?

AndrewTorrance: msg_from and survey are supplied by the client code and may not be present

PR #151, opened 2018-02-27
Refresh client token

insacuri:

Link to Trello card

**How to review**


secure_message/application.py
```diff
@@ -7,7 +7,7 @@
 from flask import json, jsonify
 from flask_restful import Api
 from flask_cors import CORS
-import redis
+import maya
```
dcdarrell9: Some other ras repos use iso8601 (if that's right) should they all be using the same datetime stuff, or is this for something else?

JamesGardiner: This is something else. They probably shouldn't be using that anyway - it's not difficult to parse a datestring using stdlib (or indeed, Maya).

secure_message/application.py
```diff
@@ -93,24 +97,18 @@ def handle_exception(error):  # NOQA pylint:disable=unused-variable
 
     return app
 
+def refresh_client_token_if_required(app):
```
AdamHawtin: The tests aren't passing because styles check picks up wrong number of blank lines here

tests/app/test_token_functions.py
```diff
@@ -1,81 +1,80 @@
 import json
```
AdamHawtin: The tests aren't passing because styles check picks up unused imports here

ajmaddaford:
[Look up users in SM (8)](https://trello.com/c/LEWKteUR/67-look-up-users-in-sm-8)

JamesGardiner:
@AndrewTorrance the expected behaviour is that the app will be resilient to failure of its backing services, and will continue to attempt to connect indefinitely (or until a SIGKILL or other stop signal is received). Ideally we would have graceful degradation of service, but that is probably for another ticket and a wider tech discussion.

secure_message/application.py
```diff
@@ -93,24 +98,18 @@ def handle_exception(error):  # NOQA pylint:disable=unused-variable
...
+    if app.token_expires_at <= maya.now():
+        cache_client_token(app)
 
-    r = redis.StrictRedis(host=app.config['REDIS_HOST'],
-                          port=app.config['REDIS_PORT'],
-                          db=app.config['REDIS_DB'])
-    put_token(r, token)
 
+def cache_client_token(app):
+    app.token = get_client_token(app.config['CLIENT_ID'],
```
warrenbailey: I think this variable would be better named oauth_client_token, token is very generic

secure_message/application.py
```diff
@@ -71,11 +72,15 @@ def create_app(config=None):
     api.add_resource(MessageSendV2, '/v2/messages')
     api.add_resource(MessageCounterV2, '/v2/messages/count')
 
+    app.token_expires_at = maya.now()
```
warrenbailey: This isn't needed as you're doing it in the cache_client_token function.

secure_message/application.py
```diff
@@ -93,24 +98,18 @@ def handle_exception(error):  # NOQA pylint:disable=unused-variable
     return app
 
 
-def cache_client_token(app):
-    token = get_client_token(app.config['CLIENT_ID'],
-                             app.config['CLIENT_SECRET'],
-                             app.config['UAA_URL'])
+def refresh_client_token_if_required(app):
+    if app.token_expires_at <= maya.now():
```
warrenbailey: again probably needs to be more descriptive

secure_message/application.py
JamesGardiner: I wanted to have the attribute set on the app at startup

JamesGardiner:
@warrenbailey the only thing i can find on it is this from the client grant section of the UAA docs: "UAA validates access tokens up to the time those tokens expire."

JamesGardiner:
@warrenbailey i've addressed your comments ✅

PR #154, opened 2018-03-01
First pass of improving logging

insacuri:
changed ''.format() to f'' to make it more readable and shorter.
Removed config item that is used in 1 place and will never change.

Link to Trello card

**How to review**


necrophonic:
Just about the PR itself - if we're using a PR template, can we fill in the appropriate sections? If they're not appropriate to the PR in question, either delete them or, better, put n/a :)

secure_message/repository/database.py
```diff
@@ -39,7 +39,7 @@ class SecureMessage(db.Model):
     def __init__(self, msg_id="", subject="", body="", thread_id="", collection_case='',
                  ru_id='', survey='', collection_exercise='', from_internal=False):
 
-        logger.debug("Initialised Secure Message entity: msg_id: {}".format(id))
+        logger.debug(f"Initialised Secure Message entity: msg_id: {id}")
```
necrophonic: [Out of scope of this PR] - I've raised an issue (https://github.com/ONSdigital/ras-secure-message/issues/155) to address that logs should be structured

config.py
```diff
@@ -81,7 +81,6 @@ class Config:
     RM_CASE_SERVICE_PORT = os.getenv('RM_CASE_SERVICE_PORT', 8171)
     RM_CASE_SERVICE_PROTOCOL = os.getenv('RM_CASE_SERVICE_PROTOCOL', 'http')
     RM_CASE_SERVICE = f'{RM_CASE_SERVICE_PROTOCOL}://{RM_CASE_SERVICE_HOST}:{RM_CASE_SERVICE_PORT}/'
-    RM_CASE_POST = '{}cases/{}/events'
```
warrenbailey: Not sure this needs to be in global app config, it could just be a constant in case_service.py

insacuri: I don't get what you mean.  It was there before and I'm removing it as it's indirection for indirections sake.  It was also used 1 time in the whole of secure message (not including the tests where it's used twice).

secure_message/repository/database.py
insacuri: Awesome, I agree that's definitely a better way of doing it.

config.py
warrenbailey: Sorry my bad, read it quickly and thought you'd added it.  You've done what I wanted anyway

PR #157, opened 2018-03-02
Improve logging again

insacuri:

Changed a number of "string".format() to f"string" to make it read easier and shorten the length of log lines.
Added additional logging to aid in debugging

**How to review**

### DIFF
--- a/tests/app/test_app.py
+++ b/tests/app/test_app.py
@@ -163,7 +163,7 @@ class FlaskTestCase(unittest.TestCase):
             self.client.post(url, data=json.dumps(test_message), headers=self.headers)
             self.assertTrue(True)  # i.e no exception
         except Exception as e:
-            self.fail("post raised unexpected exception: {0}".format(e))
+            self.fail(f"post raised unexpected exception: {e}")
 
     def test_message_post_stores_labels_correctly_for_sender_of_message(self):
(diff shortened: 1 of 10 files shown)


tests/app/test_app.py
```diff
@@ -163,7 +163,7 @@ def test_missing_thread_id_does_not_cause_exception(self):
             self.client.post(url, data=json.dumps(test_message), headers=self.headers)
             self.assertTrue(True)  # i.e no exception
         except Exception as e:
-            self.fail("post raised unexpected exception: {0}".format(e))
+            self.fail(f"post raised unexpected exception: {e}")
```
JamesGardiner: Could you log the exception using logger.exception, then just log the message text in self.fail using e.message?

insacuri:
@AndrewTorrance at the moment when you get the message list, you get a log line with the JWT encrypt, and then nothing until you get the 200 back.  Debugging this is impossible because it's like a ghost town in places

tests/app/test_app.py
JamesGardiner: @AndrewTorrance I don't think so, as the test is for not raising a generic exception.

JamesGardiner: Approving this based on the above test being refactored via a github issue.

PR #159, opened 2018-03-02
internal_user_service returns constant data when a user is defined as…

AndrewTorrance:
https://trello.com/c/AHtQQz7s

When reading messages sent by a respondent then if the msg_to was set to BRES or GROUP then there were errors as no  @msg_to was being returned to the client. This change detects these in the internal_user_service and returns default information

Changes:

Modified internal_user_service to return constant data if the actor is BRES or GROUP ( BRES needed for old messages) 
Modified internal_user_service_mock to mirror service changes
Added unit tests for this in tests.app.test_internal_user_service
Added Behavioural tests in V2 message_get and messages_get to validate expected values returned in @msg_to
Minor code refactor in messages._get_user_name
Modified test_mock_party since internal_user_service does not return same data as party and this caused some tests to fail.

### DIFF
--- a/secure_message/resources/messages.py
+++ b/secure_message/resources/messages.py
@@ -18,6 +18,7 @@ from secure_message.repository.retriever import Retriever
 from secure_message.repository.saver import Saver
 from secure_message.resources.drafts import DraftModifyById
 from secure_message.services.service_toggles import party, case_service, internal_user_service
+
 from secure_message.validation.domain import MessageSchema
 
 logger = wrap_logger(logging.getLogger(__name__))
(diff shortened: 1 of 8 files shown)


secure_message/resources/messages.py
```diff
@@ -145,18 +146,19 @@ def _inform_case_service(message):
...
-
-            if user_data:
-                first_name = user_data.get('firstName', '')
-                last_name = user_data.get('lastName', '')
-                full_name = f'{first_name} {last_name}'.strip()
-                if full_name:
-                    user_name = full_name
+            return constants.BRES_USER
+
+        user_name = 'Unknown user'
```
insacuri: Will we ever not have a username?  If we don't, what happens?  Hardcoding an 'unknown user' name seems odd to me but I explain exactly why...

PR #162, opened 2018-03-02
Feature/add filter options to get threads

AndrewTorrance:
Currently the get threads endpoint does not support the ability to filter the selection by any parameters , we need to be able to filter by survey in order that we can view the conversations related to a survey.

https://trello.com/c/gfaiP90r

This only covers the ability to filter by survey , ru etc . It does not include the ability to indicate if a thread has an unread message.

Have modified the get threads to use the message args ( survey, ru , collection case etc) 
Have cleaned up request args passing by making more functions use the MessageArgs and not rely on hard to read individual parameters 
Added tests for the same

### DIFF
--- a/secure_message/common/utilities.py
+++ b/secure_message/common/utilities.py
@@ -13,8 +13,8 @@ logger = wrap_logger(logging.getLogger(__name__))
 MessageArgs = collections.namedtuple('MessageArgs', 'string_query_args page limit ru_id survey cc label desc ce')
 
 
-def get_options(args):
-    """extract options"""
+def get_options(args, label_override=None):
+    """extract options from request , allow label to be set by caller"""
 
     string_query_args = '?'
     page = 1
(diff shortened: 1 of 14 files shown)


secure_message/common/utilities.py
```diff
@@ -41,9 +42,13 @@ def get_options(args):
     if args.get('cc'):
         cc = str(args.get('cc'))
         string_query_args = add_string_query_args(string_query_args, 'cc', args.get('cc'))
-    if args.get('label'):
-        label = str(args.get('label'))
-        string_query_args = add_string_query_args(string_query_args, 'label', args.get('label'))
+    if draft_only:
+        label = Labels.DRAFT.value
+        string_query_args = add_string_query_args(string_query_args, 'label', label)
+    else:
```
warrenbailey: why not elif?

AndrewTorrance: changed

PR #165, opened 2018-03-05
Default user details if they're not in the UAA

warrenbailey:
The ras-rm-docker-dev setup has a bootstrapped user. This user 'uaa_user' does not have a given or family name so therefore secure messaging fails with a key error.

In this scenario return a default user of ONS


benjefferies:
Ran the acceptance tests against this and it fixes it 👍

PR #167, opened 2018-03-06
fix bug in recursive function calls

JamesGardiner:
This pr addresses an issue where recursive function calls were not returning properly.

### DIFF
--- a/secure_message/application.py
+++ b/secure_message/application.py
@@ -135,9 +135,8 @@ def get_client_token(client_id, client_secret, url):
         response.raise_for_status()
 
         try:
-            return response.json()
             logger.debug("Decoding client token response json")
-            logger.debug(response.json())
+            return response.json()
         except ValueError:
             logger.exception("Failed to decode response JSON. Retrying in 10 seconds.")
             sleep(10)
--- a/secure_message/application.py
+++ b/secure_message/application.py
@@ -123,21 +123,33 @@ def get_client_token(client_id, client_secret, url):
 
     get_token_url = f'{url}/oauth/token'
 
+    s = requests.Session()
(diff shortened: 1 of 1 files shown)


secure_message/application.py
```diff
@@ -123,38 +123,41 @@ def get_client_token(client_id, client_secret, url):
...
+        logger.debug("Attempting to GET client token from UAA")
+        response = s.post(get_token_url,
+                          headers=headers,
+                          params=payload,
+                          auth=(client_id, client_secret))
         response.raise_for_status()
+
+        try:
+            return response.json()
+            logger.debug("Decoding client token response json")
```
warrenbailey: These log statements will never be called

JamesGardiner: fixed

secure_message/application.py
```diff
@@ -123,38 +123,41 @@ def get_client_token(client_id, client_secret, url):
...
             sleep(1)
-            get_client_token(client_id, client_secret, url)
+            return get_client_token(client_id, client_secret, url)
         elif 400 <= e.response.status_code < 500:
             logger.warning("Client error encountered. Shutting down.")
             sys.exit(1)
     except requests.RequestException as e:
         logger.exception(f"{e.__class__.__name__} occured while retrieving client token.")
         logger.info("Retrying client token retrieval in 10 seconds.")
         sleep(10)
```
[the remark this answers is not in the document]
warrenbailey: I agree about sleeps but in this scenario the HTTP retry will be used, its only if that fails 15 times that the sleep is invoked. 

I also think that it's out of scope for this PR, we need to fix the recursive function call.

PR #169, opened 2018-03-08
Remove dependency on python-jose and pycryptodome in favour of pyjwt

JamesGardiner:

Removes Pycrypto and python-jose dependencies, which contain security vulnerabilities, in favour of pyjwt.

### DIFF
--- a/secure_message/authentication/authenticator.py
+++ b/secure_message/authentication/authenticator.py
@@ -1,7 +1,7 @@
 import logging
 
 from flask import Response, g, current_app
-from jose import JWTError
+from jwt.exceptions import *
 from structlog import wrap_logger
 from werkzeug.exceptions import BadRequest
 
@@ -49,7 +49,11 @@ def check_jwt(token):
 
         return {'status': "ok"}
 
-    except JWTError:
+    except (InvalidTokenError, DecodeError, InvalidAlgorithmError,
+    InvalidAudienceError, ExpiredSignatureError, ImmatureSignatureError,
+    InvalidIssuedAtError, InvalidIssuerError, ExpiredSignature,
+    InvalidAudience, InvalidIssuer, MissingRequiredClaimError,
+    InvalidSignatureError):
         logger.error('Failed to decrypt or decode the JWT. Is the JWT Algorithm and Secret setup correctly?')
         return Response(response="Invalid token to access this Microservice Resource", status=400, mimetype="text/html")
     except BadRequest as e:
--- a/secure_message/authentication/authenticator.py
+++ b/secure_message/authentication/authenticator.py
@@ -1,7 +1,7 @@
 import logging
 
 from flask import Response, g, current_app
-from jwt.exceptions import *
+from jwt.exceptions import InvalidTokenError
 from structlog import wrap_logger
 from werkzeug.exceptions import BadRequest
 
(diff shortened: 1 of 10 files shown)


secure_message/authentication/jwe.py
```diff
@@ -3,6 +3,7 @@
 """
 import os
 import base64
+from functools import singledispatch
```
ajmaddaford: I can't see this being used?

secure_message/authentication/authenticator.py
```diff
@@ -49,7 +49,11 @@ def check_jwt(token):
 
         return {'status': "ok"}
 
-    except JWTError:
+    except (InvalidTokenError, DecodeError, InvalidAlgorithmError,
+    InvalidAudienceError, ExpiredSignatureError, ImmatureSignatureError,
+    InvalidIssuedAtError, InvalidIssuerError, ExpiredSignature,
+    InvalidAudience, InvalidIssuer, MissingRequiredClaimError,
+    InvalidSignatureError):
```
ajmaddaford: Although not nice, I think this is acceptable. It's explicit and we're going to replace the JWT functionality with sdc-cryptography as soon as possible.

secure_message/authentication/jwe.py
```diff
@@ -44,9 +44,18 @@ def _load_keys(self, private_key, private_key_password, public_key):
 
         self.public_key = backend.load_pem_public_key(public_key_bytes)
 
+    def _encode_token(self, token):
+        """
+        Decodes string tokens without altering bytes tokens.
```
insacuri: Might need to update docstring as this isn't decoding anything

secure_message/authentication/authenticator.py
warrenbailey: It seems pointless though, all of those exceptions inherit from InvalidTokenError, so why not just catch that one?

Also the indention is wrong and doesn't meet PEP8 standards

secure_message/authentication/jwe.py
```diff
@@ -95,12 +104,20 @@ def _load_keys(self, private_key, private_key_password, public_key):
...
+        except AttributeError:
+            return token
+
     def decrypt_token(self, encrypted_token):
         """
         Function to decrypt encrypted jwt
         """
-        if not isinstance(encrypted_token, str):
-            encrypted_token = encrypted_token.decode()
+        encrypted_token = self._decode_token(encrypted_token)
```
insacuri: 🎨 This line reads weirdly to me. Yes it's less efficient to create a new variable when we could recycle the old one, but after decoding the token then shouldn't it be 'decoded_token' or even just 'token'.  Not super important though

JamesGardiner: I'm going to keep this as encrypted_token for now as the token is still fundamentally the same but is just going from a bytes object to a str.

warrenbailey: I agree took me a while to figured out what _decode_token was actually doing, I actually preferred it before. However all of this could go if we used the crypto library

secure_message/authentication/authenticator.py
JamesGardiner: Not all of them https://github.com/jpadilla/pyjwt/blob/master/jwt/exceptions.py

warrenbailey: I disagree they all do, apart from InvalidKeyError which you don't catch anyway

PR #170, opened 2018-03-09
Make linting part of testing and enable on Travis

ajmaddaford:
This pull request enables linting on Travis. This has been done via the make targets so there is consistency between running the commands locally and on Travis.

The pipenv --style option has been deprecated (https://github.com/pypa/pipenv/blob/master/HISTORY.txt#L66) so this pull request also uses flake8 explicitly.

**How to review**

- Run make lint locally, it should run successfully
- Verify Travis runs the linting

**Note**: Travis will fail until #169 merges as the `pipenv check` command is now being run properly and finds security vulnerabilities.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,4 +11,4 @@ lint:
 	pipenv check ./secure_message ./tests
 
 test: lint
-	pipenv run run_tests.py
+	pipenv run python run_tests.py
--- a/Makefile
+++ b/Makefile
@@ -11,4 +11,4 @@ lint:
 	pipenv check ./secure_message ./tests
 
 test: lint
-    pipenv run run_tests.py
+	pipenv run run_tests.py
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -6,6 +6,9 @@ build:
...
 
-test:
-	pipenv check --style ./secure_message ./tests
-	pipenv run pytest
+lint:
+	pipenv run flake8 ./secure_message ./tests
+	pipenv check ./secure_message ./tests
+
+test: lint
+    pipenv run run_tests.py
```
dcdarrell9: I think the makefile needs to be all tabbed, otherwise we get that missing separator error in Travis

ajmaddaford: Good spot! PyCharm has converted my tab to spaces 😠

PR #171, opened 2018-03-13
Bug/internal user not seeing threads if other int user sends

AndrewTorrance:
The get threads endpoint had a bug whereby for internal users , if internal user A sent a message and internal user B got the thread list then B would not see A's sent message . In looking for this a more obscure bug was found which was if the last message on a thread was a draft sent by the respondent then all internal users would not see any conversations on the thread.

The first bug ( Int B not seeing int A's messages) was caused by the fact that the actor filtering was done on the users UUID or a constant GROUP . Which meant that any other specific UUID was not visible . The second bug was also in search caused by the use of the latest event on a thread , which could have been a draft .

To fix these the get threads was modified to look for the latest status depending on if it was sent from or to an internal user .
The label extraction code  in the message serialisation also had to be modified to account for the possibility of multiple internal UUIDs .This was changed to look for status of actors who are not the respondent.

### DIFF
--- a/secure_message/repository/database.py
+++ b/secure_message/repository/database.py
@@ -89,10 +89,14 @@ class SecureMessage(db.Model):
 
     def _populate_to_from_and_labels_internal_user(self, message):
         """populate the labels and to and from if the user is internal"""
-        if message['from_internal']:
-            respondent = [x.actor for x in self.statuses if x.label == Labels.INBOX.value or x.label == Labels.DRAFT_INBOX.value][0]
-        else:
-            respondent = [x.actor for x in self.statuses if x.label == Labels.SENT.value or x.label == Labels.DRAFT.value][0]
+        try:
+            if message['from_internal']:
+                respondent = [x.actor for x in self.statuses if x.label == Labels.INBOX.value or x.label == Labels.DRAFT_INBOX.value][0]
+            else:
+                respondent = [x.actor for x in self.statuses if x.label == Labels.SENT.value or x.label == Labels.DRAFT.value][0]
+        except IndexError:
+            logger("Could not determine respondent from message", msg_id=message["msg_id"])
+            raise
         for row in self.statuses:
             if row.actor != respondent:
                 message['labels'].append(row.label)
--- a/secure_message/repository/database.py
+++ b/secure_message/repository/database.py
@@ -103,7 +103,7 @@ class SecureMessage(db.Model):
             self._add_to_and_from(message, row)
 
     def _populate_to_from_and_labels_respondent(self, user, message):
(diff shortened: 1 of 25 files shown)


secure_message/repository/database.py
```diff
@@ -78,12 +78,49 @@ def serialize(self, user, body_summary=False):
...
+        except IndexError:
+            logger("Could not determine respondent from message", msg_id=message["msg_id"])
+            raise
+        for row in self.statuses:
+            if row.actor != respondent:
+                message['labels'].append(row.label)
+            self._add_to_and_from(message, row)
+
+    def _populate_to_from_and_labels_respondent(self, user, message):
+        """Populate labels and to and from if teh user is a respondent"""
```
JamesGardiner: Typo in docstring

secure_message/repository/database.py
```diff
@@ -78,12 +78,49 @@ def serialize(self, user, body_summary=False):
...
+            if row.actor == user.user_uuid:
+                message['labels'].append(row.label)
+            self._add_to_and_from(message, row)
+
+    @staticmethod
+    def _add_to_and_from(message, row):
+        """Populate the message to and from"""
+        if row.label == Labels.INBOX.value:
+            message['msg_to'].append(row.actor)
+        elif row.label == Labels.SENT.value:
```
warrenbailey: Could simplify this if else statement with the use of in

if row.label in [Labels.INBOX.value, Labels.DRAFT_INBOX]:
            message['msg_to'].append(row.actor)
elif row.label ins [Labels.SENT.value, Labels.DRAFT]:

secure_message/repository/retriever.py
```diff
@@ -195,6 +197,49 @@ def retrieve_thread_list(user, request_args):
...
+
+            conditions.append(SecureMessage.thread_id == t.c.thread_id)
+            conditions.append(Status.id == t.c.status_id)
+
+            result = SecureMessage.query.join(Events).join(Status) \
+                .filter(and_(*conditions)) \
+                .order_by(t.c.status_id.desc()).paginate(request_args.page, request_args.limit, False)
+
+        except Exception as e:
+            logger.exception('Error retrieving messages from database')
```
warrenbailey: Any reason for catching Exception rather than the top level SQLAlchamey exception?

secure_message/repository/database.py
AndrewTorrance: Yep , good call , changing

secure_message/repository/retriever.py
AndrewTorrance: Nope , it should be more specific . Same issue on other get list type endpoints . Changing those also.

PR #172, opened 2018-03-13
Bug/fix thread endpoint cutting off body

AdamHawtin:
**Fix a bug where the thread endpoint was only returning 100 characters of the message body**

Updated the paginated list to json function so that it accepts a keyword argument to summarize the message body (cut it down to 100 characters). It defaults to true since the messages and threads endpoints rely on this to return a summary.

**How to review:**
Run this branch of secure message locally, try sending a long message and then viewing it by the thread by ID endpoint. It should now return the entire message body.

### DIFF
--- a/tests/behavioural/features/steps/body_field.py
+++ b/tests/behavioural/features/steps/body_field.py
@@ -31,8 +31,21 @@ def step_impl_the_msg_body_is_set_too_long(context):
     context.bdd_helper.message_data['body'] = "x" * (constants.MAX_BODY_LEN + 1)
 
 
+@given("the message body is '{body_length}' characters long")
+def step_impl_the_msg_body_is_set_too_long(context, body_length):
+    """set the body to be body_length characters long"""
+    context.bdd_helper.message_data['body'] = "x" * int(body_length)
+
+
 @then("retrieved message body is as was saved")
 def step_impl_retrieved_body_is_as_saved(context):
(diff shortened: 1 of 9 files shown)


tests/behavioural/features/steps/body_field.py
```diff
@@ -31,8 +31,29 @@ def step_impl_the_msg_body_is_set_too_long(context):
     context.bdd_helper.message_data['body'] = "x" * (constants.MAX_BODY_LEN + 1)
 
 
+@given("the message body is '{body_length}' characters long")
```
AndrewTorrance: This is a duplicate function name , the test will still pass but it will cause confusion later . A separate function name is preferable . Found this by running pylint

tests/behavioural/features/steps/body_field.py
```diff
@@ -31,8 +31,29 @@ def step_impl_the_msg_body_is_set_too_long(context):
...
+
 @then("retrieved message body is as was saved")
 def step_impl_retrieved_body_is_as_saved(context):
-    """validate that the received body was teh same as was sent"""
+    """validate that the received body was the same as was sent"""
     msg_resp = json.loads(context.response.data)
     nose.tools.assert_equal(msg_resp['body'], context.bdd_helper.last_saved_message_data['body'])
+
+
+@then("the threads first message body is as was saved")
```
AndrewTorrance: Duplicate function name

PR #173, opened 2018-03-14
Enable pylint on Travis and fix lint issues

ajmaddaford:
Switched to use built in pylint mccabe over external plugin as the external plugin is not actively maintained and has issues with the latest version of the astroid module - see ONSdigital/eq-survey-runner#1469

**How to review**

Run `make lint`


secure_message/services/service_toggles.py
```diff
@@ -1,12 +1,13 @@
 import logging
+from structlog import wrap_logger
```
JamesGardiner: This should be in a separate import block

JamesGardiner: Space between structlog and logging import (stdlib vs installed)

dcdarrell9:
It would be nice if we could cut down on the output of the behave tests in secure-message as what we're getting now is pretty much not helpful

PR #174, opened 2018-03-20
Speed up tests by only creating the Flask test client once

ajmaddaford:

The behave tests were creating the Flask test client on every interaction with an endpoint, which meant the tests took far longer than necessary. This pull request creates the test client before all tests in `before_all` leading to a ~4x speedup of the tests.

This pull request also sets the config to `TestConfig` when creating the app. This allows use of the behave command to run all tests - `pipenv run behave` - or run an individual feature - `pipenv run behave tests/behavioural/features/v2/thread_get.feature` 

**How to review**

- Run all tests with `pipenv run behave`
- Run an individual feature with `pipenv run behave tests/behavioural/features/v2/thread_get.feature`

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,13 +58,30 @@ Test the application
 --------------------
 Ensure dev dependencies have been installed
 ```bash
-pipenv install --dev
+make build
 ```
 
+Ensure there is a postgres instance running on port 5432
 ```bash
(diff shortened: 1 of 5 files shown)


insacuri:
Could you add a section to the readme about running an individual behave test?  If you copied the 'how to review' section above then that would be helpful to people down the line

PR #175, opened 2018-03-20
Fix label updating to handle GROUP and internal users

AdamHawtin:
**Fix for bug in modify labels endpoint**

The endpoint to modify labels on a message still had a hardcoded BRES constant, it could not cope with messages from or to specific internal users or the GROUP constant.
I have added logic to choose the appropriate actor so that the endpoint works.

**How to review:**

Currently these fixes are only needed for features on this branch of response-operations-ui:
[sm114/distinguish-unread-messages](https://github.com/ONSdigital/response-operations-ui/tree/sm114/distinguish-unread-messages)

**To replicate the bug:**
Run that response-operations branch and open an unread message from a respondent. The call to mark messages as read will not work, when you got back to the inbox it will still be highlighted unread as the call to update the label fails.

If you this branch of secure messaging it should function as expected where opening a message will mark it as read so returning to the inbox the opened message should no longer be highlighted.

### DIFF
--- a/secure_message/repository/modifier.py
+++ b/secure_message/repository/modifier.py
@@ -15,10 +15,26 @@ logger = wrap_logger(logging.getLogger(__name__))
 class Modifier:
     """Modifies message to add / remove statuses"""
 
+
+    @staticmethod
+    def _get_label_actor(user, message):
+        try:
+            if user.is_respondent:
+                actor = user.user_uuid
+            elif message['from_internal']:
+                actor = message['msg_from']
+            else:
+                actor = message['msg_to'][0]
+        except KeyError:
+            logger.exception("Failed to remove label, no msg_to field")
+            raise InternalServerError(description="Error getting actor details from message")
+
+        return actor
+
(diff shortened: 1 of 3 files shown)


secure_message/repository/modifier.py
```diff
@@ -15,10 +15,26 @@
...
+        except KeyError:
+            logger.exception("Failed to remove label, no msg_to field")
+            raise InternalServerError(description="Error getting actor details from message")
+
+        return actor
+
     @staticmethod
     def add_label(label, message, user, session=db.session):
         """add a label to status table"""
-        actor = user.user_uuid if user.is_respondent else constants.BRES_USER
```
dcdarrell9: I think we needed to keep the constant.BRES_USER because of something to do with the old R-OPS using it? I can't remember because Torrance was working on it, but definitely something we need to check.

dcdarrell9: Discussed and should be okay, waiting for unit tests 👍

PR #176, opened 2018-03-20
Improve behave test output

dcdarrell9:

The output from the behave tests was over powering everything and very difficult to see what's going on.
This PR changes the format of the tests which shows: 
* `.` for a pass
* `F` for a fail
* `S` for a skip

*Extras*
The run_behave_tests.py doesn't work in terminal, but works in pycharm.
run_lint.py fails with `file_stream` .. This can be solved by changing pylint_mccabe to pylint.extensions.mccabe in the .pylintrc file

**How to review**

To test, run the run_tests.py to see the output, and change a feature and unit test to fail to see the results.
Try running `make test` and check the output.
Suggest improvements

### DIFF
--- a/secure_message/logger_config.py
+++ b/secure_message/logger_config.py
@@ -15,11 +15,11 @@ def logger_initial_config(service_name=None,
     if not logger_date_format:
         logger_date_format = os.getenv('LOGGING_DATE_FORMAT', "%Y-%m-%dT%H:%M%s")
     if not log_level:
-        log_level = os.getenv('SMS_LOG_LEVEL', 'INFO')
+        log_level = os.getenv('SMS_LOG_LEVEL', 'DEBUG')
     if not logger_format:
         logger_format = "%(message)s"
     if not service_name:
-        service_name = os.getenv('NAME', 'ras-frontstage')
+        service_name = os.getenv('NAME', 'ras-secure-message')
     try:
         indent = int(os.getenv('JSON_INDENT_LOGGING'))
     except TypeError:
--- a/Makefile
+++ b/Makefile
@@ -12,4 +12,5 @@ lint:
 	pipenv check ./secure_message ./tests
 
 test: lint
-	pipenv run python run_tests.py
+	pipenv run behave --format progress
+	pipenv run pytest
--- a/Makefile
+++ b/Makefile
@@ -13,4 +13,4 @@ lint:
 
 test: lint
 	pipenv run behave --format progress
-	pipenv run pytest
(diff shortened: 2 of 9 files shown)


secure_message/logger_config.py
```diff
@@ -1,44 +1,44 @@
...
-
-    logging.basicConfig(stream=sys.stdout,
-                        level=log_level,
-                        format=logger_format)
-    configure(processors=[add_log_level,
-                          filter_by_level,
-                          add_service,
-                          TimeStamper(fmt=logger_date_format, utc=True, key="created_at"),
-                          JSONRenderer(indent=indent)])
+# import logging
```
insacuri: Why is this file commented out?  Either delete it as it's not used by anything, or uncomment it out

secure_message/logger_config.py
```diff
@@ -1,44 +1,44 @@
...
+#                           logger_date_format=None):
+#
+#     if not logger_date_format:
+#         logger_date_format = os.getenv('LOGGING_DATE_FORMAT', "%Y-%m-%dT%H:%M%s")
+#     if not log_level:
+#         log_level = os.getenv('SMS_LOG_LEVEL', 'INFO')
+#     if not logger_format:
+#         logger_format = "%(message)s"
+#     if not service_name:
+#         service_name = os.getenv('NAME', 'ras-frontstage')
```
insacuri: If this file is still needed, change this default as it's wrong

dcdarrell9: That's what I was thinking, I commented it out just for testing then left it, I'm not sure this is actually needed

insacuri:
I not a fan of having all these run_X_.py files littering the place.  I think that the other speed up branch should be merged in first, then we can apply the flags that you've done to the makefile commands in there, rather then having another 2 scripts that do only subtly different things that can be done via command line

run_behave_tests.py
```diff
@@ -0,0 +1,12 @@
+import os
+import sys
+
+
+if __name__ == '__main__':
+    os.environ['APP_SETTINGS'] = 'TestConfig'
+
+    from behave import __main__ as behave_executable
+    behave_errors = behave_executable.main(args="--format progress")
```
atrozzo: would not be good to loop over the rrors and print them in a human redable way so that we can aknowledge the issues ?

secure_message/logger_config.py
insacuri: Having this commented out actually means there is no logging in the application whatsoever.  This pretty much has to be here..

PR #177, opened 2018-03-20
Fix logging

insacuri:

Link to Trello card

**How to review**


secure_message/resources/health.py
```diff
@@ -43,7 +43,6 @@ def get():
         details = {'Name': current_app.config['NAME'],
                    'Version': current_app.config['VERSION'],
                    'SMS Log level': current_app.config['SMS_LOG_LEVEL'],
-                   'APP Log Level': current_app.config['APP_LOG_LEVEL'],
```
dcdarrell9: in test_health_monitor.py I think it's looking for this, which is causing the test to fail.

Makefile
```diff
@@ -12,4 +12,5 @@ lint:
 	pipenv check ./secure_message ./tests
 
 test: lint
-	pipenv run python run_tests.py
+	pipenv run behave --format progress
+	export APP_SETTINGS=TestConfig && pipenv run python -m unittest && unset APP_SETTINGS
```
dcdarrell9: I think we could probably change this to just `pipenv run pytest` I think the run_tests.py file is a bit bloated especially when this does the same thing?

insacuri: Forgot pytest even existed... thanks ^^

secure_message/resources/health.py
insacuri: Thanks for fixing that

Makefile
```diff
@@ -12,4 +12,5 @@ lint:
 	pipenv check ./secure_message ./tests
 
 test: lint
-	pipenv run python run_tests.py
+	pipenv run behave --format progress
+	export APP_SETTINGS=TestConfig && pipenv run pytest && unset APP_SETTINGS
```
JamesGardiner: I don't think exporting this here will have any effect on the environment in which pipenv is executing the pytest script from. You'll need to use a .env for that or run the export and unset inside the pipenv shell

README.md
```diff
@@ -66,17 +66,16 @@ Ensure there is a postgres instance running on port 5432
 docker run -d -p 5432:5432 --name postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=postgres -d postgres
 ```
 
-Run them using coverage
+Run the with make
```
dcdarrell9: 🎨

PR #178, opened 2018-03-22
ModifiedRetriever so that internal user other than who sent a message…

AndrewTorrance:
Bug fix - internal user could not retrieve thread 

It was possible to see a thread in threads list but when selecting it they could not see the thread

The cause was that retriever get thread had a filter clause that status was returned only if the thread had messages where the actor was set to either  'GROUP' , 'BRES' or the UUID . In the case where internal user A sent then the UUID would be their UUID and internal user B would not be able to retrieve it . So changed it so that UUID was ignored for internal user and used status values and 'from internal' was used which is how the the get threads list does it.


warrenbailey:
Tested locally and it solves the issue

PR #179, opened 2018-03-22
Up message limit to 2k.  Added extra logging

insacuri:

Fixed a bug where a conversation of more than 10 messages gets cut off to only the 10 most recent messages

**How to review**

Send a message to respondant.  Log into frontstage as respondant, reply 10 times.  You should see 11 messages whereas before this fix you would only see 10.


secure_message/constants.py
```diff
@@ -29,7 +29,7 @@
 THREAD_LIST_ENDPOINT = "threads"
 DRAFT_LIST_ENDPOINT = "drafts"
 
-MESSAGE_QUERY_LIMIT = 20
```
dcdarrell9: How did this work when we get a 10 message limit?

warrenbailey:
Temporary fix to allow the system to move forward while we investigate and correctly fix the current pagination.

PR #182, opened 2018-03-26
Improve draft code readability

insacuri:

Renamed check_etag to something more descriptive, reversed the truthyness
of the output so we could drop the 'is false' after the check to make it simpler
to read as an english sentance.
Improved the docstring on it, and updated the unit tests to reflect the altered
functionality.

**How to review**

Save a draft in ras-frontstage.  It should work as before.  The functionality hasn't changed, just one of the methods had its truth test reversed to make it more readable.

### DIFF
--- a/secure_message/resources/drafts.py
+++ b/secure_message/resources/drafts.py
@@ -114,9 +114,7 @@ class DraftModifyById(Resource):
             logger.error('Draft put requires valid draft')
             raise BadRequest(description="Draft put requires valid draft")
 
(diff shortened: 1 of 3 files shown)


ajmaddaford:
It might be worth holding fire on reviewing this, there are some discussions about removing the draft functionality (analytics has shown it's not being used).

insacuri:
Going to remove draft functionality shortly so I'll close this

PR #183, opened 2018-03-27
Tidy get options

insacuri:

The get_options was a bit of a mess.  It seemed long and a bit overcomplicated considering what it was doing.  Only 1 place was using the string list of the arguments so that's been culled and done specifically in the one place that it's required.
Moving that allowed the get_options method to become a lot smaller.

**How to review**

Everything should function as before, try to get a message, or a conversation and it should work as expected.

### DIFF
--- a/secure_message/common/utilities.py
+++ b/secure_message/common/utilities.py
@@ -1,6 +1,7 @@
 import collections
 import hashlib
 import logging
+import urllib.parse
 
 from structlog import wrap_logger
 from secure_message.common.labels import Labels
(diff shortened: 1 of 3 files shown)


secure_message/common/utilities.py
```diff
@@ -8,85 +8,67 @@
...
+    string_query = "?"
+    for field in args._fields:
+        if field in ['page']:
+            pass
+        value = getattr(args, field)
+        if value is not None:
+            if string_query == '?':
+                string_query = string_query + f'{field}={value}'
+            else:
+                string_query = string_query + f'&{field}={value}'
```
ajmaddaford: The query string should be generated using `urllib.parse.urlencode` rather than string concatenation (which is error prone and less clear) - see https://docs.python.org/3.6/library/urllib.request.html#urllib-examples 

The `urlencode` method can take a dict or a sequence of two element tuples, so with some changes to how we deal with `MessageArgs` we should be able to remove this method.

secure_message/common/utilities.py
```diff
@@ -8,85 +8,67 @@
...
+    return string_query
 
 
 def process_paginated_list(paginated_list, host_url, user, message_args, endpoint=MESSAGE_LIST_ENDPOINT, body_summary=True):
```
AndrewTorrance: The endpoint defaults to the v1 get message list endpoint . Should it go to /v2/messages ?

insacuri: What's the best way to do it?  Update the constant to v2/messages?, update it in the string below?  Go to the place where process paginated list is called and tweak it there?

PR #189, opened 2018-04-23
removed bres references in the code, renamed functions and acceptance.

atrozzo:
Not been able to work on this code since my machine gave up last week. Not sure about the quality and correctness of the code. Please feel free to collect the code and chnage it .


tests/app/test_party_service.py
```diff
@@ -109,7 +109,7 @@ def test_get_user_details_for_bres_user(self, mock_request):
         mock_request.get(user_data_url, status_code=200, reason="OK", text=str(expected_result))
 
         with self.app.app_context():
-            result_data = sut.get_user_details(constants.BRES_USER)
+            result_data = sut.get_user_details(constants.NON_SPECIFIC_INTERNAL_USER)
```
ajmaddaford: Looking at the `get_user_details` method it looks like the party service was never called for the `BRES_USER` or `NON_SPECIFIC_INTERNAL_USER`. If so, now would be a good time to remove the functionality from the secure message party service code and then this test wouldn't be needed.

tests/app/test_retriever.py
```diff
@@ -143,7 +143,7 @@ def create_threads(self, no_of_threads=1, add_internal_draft=False, add_responde
             msg_id = str(uuid.uuid4())
             thread_id = msg_id
             threads.append(thread_id)
-            self.add_secure_message(msg_id=msg_id, thread_id=thread_id, survey=constants.BRES_USER, from_internal=False)
+            self.add_secure_message(msg_id=msg_id, thread_id=thread_id, survey=constants.NON_SPECIFIC_INTERNAL_USER, from_internal=False)
```
ajmaddaford: The survey should be a survey id rather than a user, it looks like this has always been wrong as the user was `BRES` which looks like a survey identifier.

tests/behavioural/features/draft_put.feature
```diff
@@ -23,7 +23,7 @@ Feature: Draft Put Endpoint
     Then retrieved message body is as was saved
 
    Scenario: A Respondent saves and edits a draft with an apostraphe
-    Given  sending from respondent to internal bres user
+    Given  sending from internal user to respondent
```
ajmaddaford: This was previously respondent to internal, should this have been switched to internal to respondent? This is the same for other occurrences below (there are a lot of them).

tests/behavioural/features/steps/to_field.py
```diff
@@ -94,10 +94,10 @@ def step_impl_the_at_msg_to_is_set_to_internal_group_user(context):
...
 
 
-@then("the at_msg_to is set correctly for bres user")
-def step_impl_the_at_msg_to_is_set_to_bres_user(context):
+@then("the at_msg_to is set correctly for internal user")
+def step_impl_the_at_msg_to_is_set_to_not_specific_user(context):
     msg_resp = json.loads(context.response.data)
-    expected = {"id": BRES_USER,
+    expected = {"id": NON_SPECIFIC_INTERNAL_USER,
                 "firstName": "BRES",
```
ajmaddaford: Should the firstName be `BRES`?

PR #193, opened 2018-04-30
Remove message by id endpoint

insacuri:

https://trello.com/c/rYMPWFRQ/164-migrate-v2-messages-functionality-to-messages-5

Part 3 of the migration away from v2 endpoints.  This PR is fairly hefty because a very large number of the unit tests used the msg_id returned after a message send to ensure that the message was saved as expected.  Now we're having to get threads and query the message within the thread.
The actual code changes themselves are minimal as we're just removing an endpoint.

**How to review**

The unit and acceptance tests should pass as expected as this is removing unused functionality.


secure_message/resources/messages.py
```diff
@@ -138,26 +136,6 @@ def _get_user_name(user, message):
...
-        message = add_users_and_business_details([message])[0]
-        if Authorizer().can_user_view_message(g.user, message):
-            return jsonify(message)
-        result = jsonify({'status': 'error'})
-        result.status_code = 403
-        logger.error('Error getting message by ID', msg_id=message_id, status_code=result.status_code)
-        return result
-
-
 class MessageModifyById(Resource):
```
MIslam92: If your removing the use of `MessageById` there would be no need for this either? or is that for the next iteration of this?

insacuri: We still need message modify by id because we need to change the labels (read -> unread for example)

PR #198, opened 2018-05-02
Remove draft

insacuri:
~~NOTE: Putting this up for early review.  Some of the unit tests are still failing but this can be reviewed now.  I wanted to start the PR so I can see how happy codecov/codacy/travis is with this~~

https://trello.com/c/WdR5QK3X/167-remove-all-remaining-draft-functionality-from-secure-message-3

All of the draft functionality is unused and doesn't need to be maintained anymore.  Removing this will result in a smaller secure-message service with less parts to maintain.

**How to review**

Unit, behave and acceptance tests should continue to work.  Everything removed was unused code so this should be reasonably low risk


secure_message/repository/retriever.py
```diff
@@ -236,9 +120,9 @@ def _retrieve_respondent_thread_list(request_args, user):
             t = db.session.query(SecureMessage.thread_id, func.max(SecureMessage.id)  # pylint:disable=no-member
                                  .label('max_id')) \
                 .join(Events).join(Status) \
-                .filter(Status.label != Labels.DRAFT_INBOX.value) \
+                .filter(Status.label != "DRAFT_INBOX") \
```
dcdarrell9: Do we need this `DRAFT_INBOX` filter?

secure_message/authorization/authorizer.py
```diff
@@ -24,7 +24,7 @@ def can_user_view_message(user, message):
         if user.user_uuid == message['msg_from']:
             return True
 
-        if user.user_uuid in message['msg_to'] and 'DRAFT_INBOX' not in message['labels']:
+        if user.user_uuid in message['msg_to']:
```
warrenbailey: the whole of this could be simplified for to:

return user.is_internal or user.user_uuid in [message['msg_from'] + message['msg_to']]

Only styinge though, so feel free to reject

rjweeks-securemessage-1.0.0-swagger.yaml
```diff
@@ -22,7 +22,7 @@ securityDefinitions:
     scopes:
       write: allows modifying resources
       read: allows reading resources
- 
+
```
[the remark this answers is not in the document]
warrenbailey: Yeah it's a pointless file now

secure_message/common/utilities.py
```diff
@@ -31,8 +30,6 @@ def get_options(args, draft_only=False):
 
     if args.get('desc') and args.get('desc') == 'false':
         fields['desc'] = False
```
warrenbailey: Why is this doing 'and' on the same function call, could do

fields['desc'] = bool(strtobool(args.get('desc'))

secure_message/repository/retriever.py
warrenbailey: Agreed, that should be removed

PR #199, opened 2018-05-09
Performance testing

insacuri:

https://trello.com/c/jSBCML7Y/182-simple-performance-test-for-message-listing-5

During the performance testing, there were a few files created that could be reused in the future.  They were moved into a sensible place and a readme written on roughly how to use them for the next person to come along.

**How to review**

N/A, it's non functional so just make sure the code and readme read correctly and are useful.


tests/performance/README.md
```diff
@@ -0,0 +1,37 @@
...
+- Install locust, `make build` should do it, but if it's not in the pipfile then `pipenv install locustio`
+- `pipenv run locust -f /path/to/locustfile.py` - this will start a server on `http://localhost:8089` that uses the specified locustfile
+-  Go to the above url, specify the number of users and the spawn rate
+
+#### create-messages script guide
+To create a number of messages quickly, there is a create-messages.py script.
+The problem with this script is that it needs correct ids for it to work.  To get these ids,
+follow the steps below.
+
+- Run acceptance tests to get test data into response-operations-ui
```
dcdarrell9: Might be worth saying that docker needs to be running

tests/performance/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+
+### Locust
+
+- Install locust, `make build` should do it, but if it's not in the pipfile then `pipenv install locustio`
+- `pipenv run locust -f /path/to/locustfile.py` - this will start a server on `http://localhost:8089` that uses the specified locustfile
+-  Go to the above url, specify the number of users and the spawn rate
+
+#### create-messages script guide
+To create a number of messages quickly, there is a create-messages.py script.
```
dcdarrell9: If this needs to be done first it might be handy to write a `run_perforamce_test.sh` which gets everything set up so that it can be run (steps below etc)

Pipfile
```diff
@@ -27,10 +22,9 @@ requests = "*"
 retrying = "*"
 maya = "*"
 PyJWT = "*"
-
+locustio = "*"
```
warrenbailey: I think this should be in the dev packages

dcdarrell9:
Looking good 👍 

100 users:
Average time on the three tests: all just over a second 👍 

150 users:
Average time on the three tests: all around 1.7 seconds 👍 

200 users:
Some failures ~ 2%
Average time on the three tests: all around 7.5 - 8 seconds 👎 

300 users:
Some failures ~ 7%
Average time on the three tests: Floating around 9.5 seconds 👎

tests/performance/create-messages.py
```diff
@@ -0,0 +1,27 @@
...
+        "ru_id": "47e5859b-5d18-4c73-a91c-05491f6c6167",
+        "collection_case": "",
+        "survey": "02b9c366-7397-42f7-942a-76dc5876d86d"
+    }
+    response = post("http://localhost:5050/v2/messages", headers={'Authorization': '<insert jwt here>',
+                    'Content-Type': 'application/json', 'Accept': 'application/json'}, data=json.dumps(message))
+
+    try:
+        response.raise_for_status()
+        print("Message {x} sucessfully sent")
```
dcdarrell9: `f"Message {x} success` at the moment comes through as is

tests/performance/README.md
insacuri: Whatcha mean sorry?  The steps to get the right data is manual so I can't script it up, and getting into a stage where it can be automated will take a bunch more work that probably isn't worth it in light of all the other stuff that we need to get done.

dcdarrell9: Yeah, I think I was tired when I wrote this, I think my idea behind it was more to do with automation but there's no need for that now, but if we could get automation in the future it would give us a lot of confidence making changes. It works now so it's not a problem 👍

PR #200, opened 2018-05-10
Remove user name from logs

dcdarrell9:

Removed a user name from the logs

**How to review**

Test the log does not print out a user name

### DIFF
--- a/secure_message/services/case_service.py
+++ b/secure_message/services/case_service.py
@@ -22,6 +22,6 @@ class CaseService:
                          reason=case_service_data.reason, text=case_service_data.text)
             return case_service_data.status_code
 
-        err = f"No case id for case involving user {user_name}, case event not called"
+        err = "No case id for case involving user, case event not called"
         logger.error(err)
         return {'error': err}, 400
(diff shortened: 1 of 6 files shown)


secure_message/services/case_service.py
```diff
@@ -22,6 +22,6 @@ def store_case_event(case_id, user_name):
                          reason=case_service_data.reason, text=case_service_data.text)
             return case_service_data.status_code
 
-        err = f"No case id for case involving user {user_name}, case event not called"
+        err = "No case id for case involving user, case event not called"
```
warrenbailey: I think we should add msg_id to the function arguments, at least then we can log the error with that id and trace it back to the message and hence user it relates to.

Also not entirely happy with that if statement, what happens if case_id is None?

dcdarrell9: So, I've been debugging through to see what's going on, and every message I've sent so far doesn't get passed this if because there's no case_id, I'm not sure where we get/use the case_id, but from what I've seen so far, it's not being used in messaging

PR #201, opened 2018-05-10
Remove jwe

insacuri:

https://trello.com/c/ribJzayk/173-remove-unused-jwt-encryption-code-and-tests-3

The JWT encryption that was written was never used so it should be deleted.  This PR removes this code and updates the tests to reflect it.

**How to review**

Everything should function as normal, unit and acceptance tests should pass.  It would probably be a good idea to send a message, read a thread list and look at a thread to double check that everything is okay as is.


API.md
```diff
@@ -68,8 +68,6 @@ this currently has two fields :
 * party_id which is how the user is identified, this should be the users uuid
 * role which is set to either 'internal' for ons staff or 'respondent' for respondents.
 
-The JWT usage does cause some issues in that currently it may or may not be encrypted. The secure message api relies on an environment variable called 'SM_JWT_ENCRYPT'. If this is set to 1 then it assumes the JWT is encrypted and attempts to decrypt it. If '0' then it skips the decryption. It is a common error to have this set incorrectly.
-
 After possible decryption, the service attempts to decode the JWT data . For that it uses an algorithm defined in config and a secret also defined in config. If the algorithm and/or secret are out of step between client and secure message service then the JWT will fail checks and the service will return a 500.
```
warrenbailey: Remove "After possible decryption"

warrenbailey:
Manually tested and it all works for me

API.md
insacuri: Fixed

PR #203, opened 2018-05-24
Add event data to messages

insacuri:

https://trello.com/c/CNzvIEvw/222-add-event-data-columns-to-securemessage-table-3

Added 2 new columns to secure_message table that were represented by the events table.  Now writing the same data to both the events table and these new columns

**How to review**

Unit + acceptance tests still pass.  This PR only adds data that isn't being used (for now) so testing this should be very very simple.


secure_message/repository/modifier.py
```diff
@@ -76,11 +78,21 @@ def add_unread(message, user):
...
+    def del_unread(message, user, session=db.session):
         """Remove unread label from status"""
         inbox = Labels.INBOX.value
         unread = Labels.UNREAD.value
         if inbox in message['labels'] and unread in message['labels'] and 'read_date' not in message:
+            # Save to both events and secure_message table for now.  In future, the save to the
+            # events table will be removed.
             Saver().save_msg_event(message['msg_id'], EventsApi.READ.value)
+            try:
+                result = SecureMessage.query.filter(SecureMessage.msg_id == message['msg_id'])
```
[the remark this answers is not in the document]
warrenbailey: I think a problem can arise here. The call to Saver().save_msg_event() commits the database session so there's a chance that the database will end up in an inconsistent state if the update to the message table fails.

Ideally the whole of this methods needs to be in a single database transaction, so I think we need to rethink this.

secure_message/repository/modifier.py
```diff
@@ -76,11 +78,21 @@ def add_unread(message, user):
...
+                result.one().read_datetime = datetime.now(timezone.utc)
+                session.commit()
+            except SQLAlchemyError:
+                session.rollback()
+                logger.exception('Error adding read_datetime to message', msg_id=message['msg_id'])
+                raise InternalServerError(description="Error adding read_datetime to message")
```
warrenbailey: and interestingly the remove_label below doesn't commit the db session at all

migrations/versions/b1fdd85b9ca3_add_columns_to_securemessage.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def upgrade():
+    op.add_column('secure_message', sa.Column('sent_datetime', sa.DateTime), schema='securemessage')
+    op.add_column('secure_message', sa.Column('read_datetime', sa.DateTime), schema='securemessage')
+
+
+def downgrade():
+    op.drop_column('secure_message', 'sent_datetime', schema='securemessage')
+    op.drop_column('secure_message', 'read_datetime', schema='securemessage')
```
ajmaddaford: 🎨 Personal preference, I always prefer using _at for datetime columns e.g. `sent_at` (and `_on` for date columns e.g. `closed_on`).

insacuri: so, `sent_at` and `read_on` for column names?  I think I only went with those names because another column had a similar name, but I was never really attached to that naming convention

insacuri:
By this, the best thing to do would be to execute the sql directly.  That's fine by me, but I don't have that sort of access.   Should I write a script (should be short) in this PR and have someone else execute it?

migrations/versions/b1fdd85b9ca3_add_columns_to_securemessage.py
warrenbailey: I think the sent_at and read_at sound ok, they are descriptive.

warrenbailey:
I agree we should remove albemic, look at some agnostic DB migration tools we can run as part of the deployment pipeline. I think for now for this simple migration we just script it in SQL.

ajmaddaford:
I've raised a pull request for the database update in pure SQL - https://github.com/ONSdigital/ras-secure-message/pull/205

secure_message/repository/database.py
```diff
@@ -49,6 +51,8 @@ def __init__(self, msg_id="", subject="", body="", thread_id="", collection_case
         self.survey = survey
         self.collection_exercise = collection_exercise
         self.from_internal = from_internal
+        self.sent_at = datetime.now(timezone.utc)
```
ajmaddaford: This can be done as part of the column definition:
```
sent_at = Column('sent_at', DateTime, default=datetime.datetime.utcnow)
```

http://docs.sqlalchemy.org/en/latest/core/defaults.html

secure_message/repository/modifier.py
```diff
@@ -76,11 +77,23 @@ def add_unread(message, user):
         return res
 
     @staticmethod
-    def del_unread(message, user):
+    def del_unread(message, user, session=db.session):
```
ajmaddaford: Is there a reason to make session a parameter if it's never passed in?
Also, it might be worth renaming the method now.

secure_message/repository/modifier.py
```diff
@@ -76,11 +77,23 @@ def add_unread(message, user):
...
         inbox = Labels.INBOX.value
         unread = Labels.UNREAD.value
         if inbox in message['labels'] and unread in message['labels'] and 'read_date' not in message:
-            Saver().save_msg_event(message['msg_id'], EventsApi.READ.value)
+            # Save to both events and secure_message table for now.  In future, the save to the
+            # events table will be removed.
+            try:
+                event = Events(msg_id=message['msg_id'], event=EventsApi.READ.value)
+                session.add(event)
+                result = SecureMessage.query.filter(SecureMessage.msg_id == message['msg_id']).one()
```
ajmaddaford: Why `result` and not something more descriptive e.g. `secure_message`?

insacuri: Changed it to message

secure_message/repository/modifier.py
insacuri: Renamed the method to something more descriptive and accurate.  
I tried moving session into the function, using db.session directly and other stuff but pylint complains that with a bunch of `Instance of 'scoped_session' has no 'add' member`.  Not sure what to do to fix that, any thoughts?

secure_message/repository/database.py
insacuri: Done, much tidier ^^

secure_message/repository/modifier.py
ajmaddaford: Survey runner disables the

ajmaddaford: Another option could be to add `scoped_session` to the `ignored-classes` property in .pylintrc (see https://stackoverflow.com/questions/28193025/pylint-cant-find-sqlalchemy-query-member)

secure_message/repository/modifier.py
```diff
@@ -76,11 +77,23 @@ def add_unread(message, user):
...
+                result.read_at = datetime.now(timezone.utc)
```
ajmaddaford: Does this give anything different to `datetime.utcnow`?

insacuri: different if you look at the whole string, but it's the same when it gets saved to the database.  Changed it to that.

secure_message/repository/modifier.py
insacuri: actually changed now

PR #204, opened 2018-05-31
Close conversation

insacuri:

https://trello.com/c/RTqYEtQy/201-sm123-int-close-a-conversation-8

This PR allows users to close conversations.  This PR still needs tests written against it so it can't be merged straight away, but I'll put it up so that we can make sure that it's the right approach.

**How to review**

You'll need your database to be in the correct configuration.  This can be done by deleting the securemessage schema then starting the app.  The app will create the schema and tables for you without any intervention.

### DIFF
--- a/secure_message/repository/retriever.py
+++ b/secure_message/repository/retriever.py
@@ -3,13 +3,14 @@ import logging
 from flask import jsonify
 from sqlalchemy import and_, func, or_
 from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 from structlog import wrap_logger
 from werkzeug.exceptions import InternalServerError, NotFound
 
 from secure_message import constants
 from secure_message.common.eventsapi import EventsApi
 from secure_message.common.labels import Labels
-from secure_message.repository.database import db, Events, SecureMessage, Status
+from secure_message.repository.database import db, Conversation, Events, SecureMessage, Status
 
 logger = wrap_logger(logging.getLogger(__name__))
 
@@ -241,6 +242,18 @@ class Retriever:
             raise InternalServerError(description="Error retrieving conversation from database")
 
         return result
+
+    @staticmethod
+    def retrieve_thread_metadata(thread_id):
+        result = Conversation.query.filter(Conversation.thread_id == thread_id)
+        try:
+            return result.one()
+        except NoResultFound:
+            logger.error("No conversation found", thread_id=thread_id)
+            raise NotFound(description=f"Conversation with thread_id '{thread_id}' does not exist")
+        except MultipleResultsFound:
+            logger.error("Mulitple results found for conversation", thread_id=thread_id)
+            raise
 
     @staticmethod
     def check_db_connection():
--- a/secure_message/repository/retriever.py
+++ b/secure_message/repository/retriever.py
@@ -156,11 +156,16 @@ class Retriever:
                                  .label('max_id')) \
                 .join(Events).join(Status) \
                 .filter(or_(and_(SecureMessage.from_internal.is_(False), Status.label == Labels.INBOX.value),  # NOQA
-                            and_(SecureMessage.from_internal.is_(True),
-                                 Status.label.in_([Labels.SENT.value]))
-                           )
-                       ) \
-                .group_by(SecureMessage.thread_id).subquery('t')
+                            and_(SecureMessage.from_internal.is_(True), Status.label == Labels.SENT.value))
+                        ).group_by(SecureMessage.thread_id).subquery('t')
+
+            u = db.session.query(SecureMessage.thread_id).join(Conversation) \
+                .filter(Conversation.is_closed.is_(True)).subquery('u')
+
+            if request_args.label == 'CLOSED':
+                conditions.append(SecureMessage.thread_id.in_(u))
+            else:
+                conditions.append(~SecureMessage.thread_id.in_(u))
 
             conditions.append(SecureMessage.thread_id == t.c.thread_id)
             conditions.append(SecureMessage.id == t.c.max_id)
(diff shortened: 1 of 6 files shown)


secure_message/resources/threads.py
```diff
@@ -20,13 +23,68 @@ def get(thread_id):
...
         logger.info("Successfully retrieved messages from thread", thread_id=thread_id, user_uuid=g.user.user_uuid)
         messages = []
         for message in conversation.all():
             msg = message.serialize(g.user, body_summary=False)
             messages.append(msg)
-        return jsonify({"messages": add_users_and_business_details(messages)})
+
+        if conversation_metadata.is_closed:
+            return jsonify({"messages": add_users_and_business_details(messages),
+                            "metadata": ThreadById._build_metadata(conversation_metadata)})
```
ajmaddaford: Is there a reason for nesting this under a metadata property? An alternative could be:
```
{
    "is_closed": true,
    "closed_by": "someone",
    ...
    "messages": [
		{},
		{}
	]
}
```

secure_message/repository/retriever.py
```diff
@@ -242,6 +248,18 @@ def _retrieve_thread_for_internal_user(thread_id):
...
         return result
 
+    @staticmethod
+    def retrieve_thread_metadata(thread_id):
+        result = Conversation.query.filter(Conversation.thread_id == thread_id)
+        try:
+            return result.one()
+        except NoResultFound:
+            logger.error("No conversation found", thread_id=thread_id)
+            raise NotFound(description=f"Conversation with thread_id '{thread_id}' does not exist")
```
ajmaddaford: A conversation record won't always exist when we first deploy these changes (we'll need to create conversation records for all existing conversations after we've deployed this), so I don't think we should class no conversation record as an error.

secure_message/resources/threads.py
```diff
@@ -20,13 +23,68 @@ def get(thread_id):
...
+            "closed_by_uuid": conversation_metadata.closed_by_uuid,
+            "closed_on": conversation_metadata.closed_on
+        }
+
+    @staticmethod
+    def patch(thread_id):
+        """Modify conversation metadata"""
+
+        logger.info("Attempting to modify metadata for thread", thread_id=thread_id, user_uuid=g.user.user_uuid)
+        request_data = request.form
```
ajmaddaford: `request.form` should be `request.json`

secure_message/resources/threads.py
```diff
@@ -20,13 +23,68 @@ def get(thread_id):
...
+        msg_property, value = ThreadById._validate_request(request_data)
+
+        logger.info("Getting metadata for thread", thread_id=thread_id, user_uuid=g.user.user_uuid)
+        metadata = Retriever.retrieve_thread_metadata(thread_id)
+        if msg_property == 'is_closed':
+            if value:
+                if metadata.is_closed:
+                    logger.info("Conversation already closed", thread_id=thread_id, user_uuid=g.user.user_uuid)
```
ajmaddaford: Do we even care what the request data is if the conversation is closed? If not we could make this the first thing we check.

secure_message/resources/threads.py
```diff
@@ -20,13 +23,68 @@ def get(thread_id):
...
+    def _validate_request(request_data):
+        """Used to validate data within request body for ModifyById"""
+        if not g.user.is_internal:
+            logger.info("Thread modification is forbidden")
+            abort(403)
+        # Check if it's empty
+        if not request_data:
+            logger.error('No properties provided')
+            raise BadRequest(description="No properties provided")
+        if request_data['is_closed'] not in ['True', 'False']:
```
ajmaddaford: These shouldn't be strings, they should be JSON booleans (see the comment about `request.json` above).

secure_message/resources/messages.py
```diff
@@ -39,6 +39,11 @@ def post(self):
         message = self._validate_post_data(post_data)
 
         if message.errors == {}:
+            if message.thread_id:
+                conversation_metadata = Retriever.retrieve_thread_metadata(message.thread_id)
```
ajmaddaford: I think we need to be clear in our use of conversation vs thread. I think anything related to the new table should refer to conversation rather than thread - so this would be `retrieve_conversation_metadata`. This also applies to all the other places this occurs in this pull request.

secure_message/repository/retriever.py
```diff
@@ -155,11 +156,16 @@ def _retrieve_internal_thread_list(request_args, user):
...
+                            and_(SecureMessage.from_internal.is_(True), Status.label == Labels.SENT.value))
+                        ).group_by(SecureMessage.thread_id).subquery('t')
+
+            u = db.session.query(SecureMessage.thread_id).join(Conversation) \
+                .filter(Conversation.is_closed.is_(True)).subquery('u')
+
+            if request_args.label == 'CLOSED':
+                conditions.append(SecureMessage.thread_id.in_(u))
+            else:
+                conditions.append(~SecureMessage.thread_id.in_(u))
```
ajmaddaford: I think this change should be done as part of the change for the listing. If we do this without updating the frontend, won't the closed messages disappear from the listing?

secure_message/resources/threads.py
insacuri: Oddly enough that was my first design, should've trusted my gut on that ^^

secure_message/repository/retriever.py
insacuri: Agreed, I thought the same thing and thought I'd changed it to that.  Must've changed it somewhere else sorry >_<

secure_message/repository/retriever.py
insacuri: Thats fair.  I'll make a closed-conversation-only branch that won't have these changes in it

insacuri:
Closing because a more finished PR is up

PR #205, opened 2018-06-01
Add migration for event date columns on message table

ajmaddaford:

Adds `sent_at` and `read_at` columns to the `secure_message` table in preparation for the code change that will populate these columns.

**How to review**

- Run the migration SQL against the secure message database, if running the ras-rm-docker-dev setup you can use `psql -h localhost -p 6432 -U postgres -d postgres -a -f migrations/001_create_event_dates_on_message.sql`
- Use the service and verify it works
- Run the acceptance tests


ajmaddaford:
@insacuri They are the defaults and are the same as the `date_time` column on events created by SQLAlchemy.

PR #206, opened 2018-06-05
Fix pagination bug

gavinleeedwards:

https://trello.com/c/q17Lr227/208-fix-pagination-in-secure-messaging

A bug existed in secure message where not all pagination links were showing.  When retrieving the counts of messages by survey, the SQL was retrieving a count of any conversations logged against the current user (their uuid) or GROUP.  If  different user logged in then the count wasn't retrieving all conversations.  
&nbsp;  
&nbsp;  
**How to review**

To recreate the issue:

- Bring everything up in Docker
- Run the acceptance tests to set up conversations
- Log in as uaa_user and create a few more conversations to push the pagination to the 5th page
- Log in as a different internal user (see below for instructions)
- Navigate to messages - only the first 2 pages are displayed, although you can change the page number in the URL to navigate to the 4th and 5th pages.

- Run this branch locally and navigate to messages (both as uaa_user and a different user).  The same number of messages should be displayed for each.
- Acceptance tests should still all pass

&nbsp;  
&nbsp;  

**To create a new internal user through UAA**

- Retrieve a bearer token:

`curl 'http://localhost:9080/oauth/token' -i -u 'admin:admin_secret' -X POST -H 'Content-Type: application/x-www-form-urlencoded' -H 'Accept: application/json' -d 'grant_type=client_credentials&response_type=token&token_format=opaque'`

- Create a new user, replacing the bearer token with the one you've retrieved and adding in your details (username, name, password etc.):

`curl 'http://localhost:9080/Users' -i -X POST -H 'Accept: application/json' -H 'Authorization: Bearer 34224c536b3648bd8d166acc5be4159f' -H 'Content-Type: application/json'  -d '{
"userName" : "your_username",
"name" : {
"formatted" : "given name family name",
"familyName" : "Surname",
"givenName" : "First_name"
},
"emails": [{
"value": "email@email.com",
"primary": true
}],
"active": true,
"verified": true,
"password": "password"
}’`

### DIFF
--- a/secure_message/application.py
+++ b/secure_message/application.py
@@ -22,8 +22,8 @@ from secure_message.logger_config import logger_initial_config
 from secure_message.repository import database
 from secure_message.resources.health import DatabaseHealth, Health, HealthDetails
 from secure_message.resources.info import Info
-from secure_message.resources.messages import MessageCounter, MessageModifyById, MessageSend
-from secure_message.resources.threads import ThreadById, ThreadList
+from secure_message.resources.messages import MessageModifyById, MessageSend
+from secure_message.resources.threads import ThreadById, ThreadCounter, ThreadList
 
 logger = wrap_logger(logging.getLogger(__name__))
 
@@ -56,10 +56,10 @@ def create_app(config=None):
     api.add_resource(MessageSend, '/message/send', '/v2/messages')
     api.add_resource(MessageModifyById, '/message/<message_id>/modify',
                      '/v2/messages/modify/<message_id>')
-    api.add_resource(MessageCounter, '/v2/messages/count')
 
     api.add_resource(ThreadList, '/threads')
     api.add_resource(ThreadById, '/thread/<thread_id>', '/v2/threads/<thread_id>')
+    api.add_resource(ThreadCounter, '/v2/messages/count')
 
     app.oauth_client_token_expires_at = maya.now()
 
--- a/secure_message/repository/retriever.py
+++ b/secure_message/repository/retriever.py
@@ -6,7 +6,6 @@ from sqlalchemy.exc import SQLAlchemyError
 from structlog import wrap_logger
 from werkzeug.exceptions import InternalServerError, NotFound
 
-from secure_message import constants
 from secure_message.common.eventsapi import EventsApi
 from secure_message.common.labels import Labels
 from secure_message.repository.database import db, Events, SecureMessage, Status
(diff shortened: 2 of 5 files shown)


secure_message/repository/retriever.py
```diff
@@ -33,10 +33,18 @@ def unread_message_count(user):
...
+        survey_conditions = []
+
+        if survey:
+            survey_conditions.append(SecureMessage.survey.in_(survey))
+        else:
+            survey_conditions.append(True)
+
         if user.is_internal:
-            status_conditions, survey_conditions = Retriever._get_conditions_internal_user(survey, user)
+            status_conditions = []
```
MIslam92: Could you not set the status conditions like you have survey_conditions prior and reduce the need for the else? i.e. `if !user.is_internal: status_conditions = [Status.actor == str(user.user_uuid)]`

secure_message/application.py
```diff
@@ -56,10 +56,11 @@ def create_app(config=None):
     api.add_resource(MessageSend, '/message/send', '/v2/messages')
     api.add_resource(MessageModifyById, '/message/<message_id>/modify',
                      '/v2/messages/modify/<message_id>')
-    api.add_resource(MessageCounter, '/v2/messages/count')
 
     api.add_resource(ThreadList, '/threads')
     api.add_resource(ThreadById, '/thread/<thread_id>', '/v2/threads/<thread_id>')
+    api.add_resource(ThreadCounter, '/v2/messages/count',
+                     '/v2/threads/count')
```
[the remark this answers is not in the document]
dcdarrell9: If the endpoint changed wouldn't we need to change r-ops? The call for count there is hitting `/v2/messages/count`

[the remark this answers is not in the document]
gavinleeedwards: r-ops would need to be changed but this was just to add the new endpoint alongside the current one so that both would use ThreadCount until r-ops was changed.  As it stands, I think the new endpoint should be removed (as @ajmaddaford pointed out the conflict) and any further refactoring should be separate from fixing the original bug.

ajmaddaford:
We should have a test to ensure that we don't re-introduce this bug.

tests/app/test_retriever.py
```diff
@@ -396,6 +398,16 @@ def test_latest_message_from_each_thread_chosen_desc(self):
...
                     self.assertEqual(msg_ids[x], thread.all()[0].events[0].msg_id)
 
+    def test_thread_count_by_survey(self):
+        """checks that the returned thread count is the same for every internal user"""
+        self.create_threads(5)
+
+        with self.app.app_context():
+            with current_app.test_request_context():
+                thread_count_internal = Retriever().thread_count_by_survey(self.user_internal, BRES_SURVEY)
+                thread_count_second_internal = Retriever().thread_count_by_survey(self.user_internal, BRES_SURVEY)
```
LJBabbage: Shouldn't this be self.second_user_internal?

gavinleeedwards: Yep, it really should be.  Changing it now.

PR #207, opened 2018-06-05
Close conversation only

insacuri:
What is the context of this PR?

trello.com/c/RTqYEtQy/201-sm123-int-close-a-conversation-8

This PR allows users to close conversations. This PR still needs tests written against it but the code can be reviewed.

How to review

You'll need your database to be in the correct configuration. This can be done by deleting the securemessage schema then starting the app. The app will create the schema and tables for you without any intervention.

### DIFF
--- a/secure_message/repository/database.py
+++ b/secure_message/repository/database.py
@@ -137,7 +137,7 @@ class Status(db.Model):
     """Label Assignment table model"""
     __tablename__ = "status"
 
-    id = Column('id', Integer(), primary_key=True)
+    id = Column('id', Integer(), primary_key=True, index=True)
     label = Column('label', String(constants.MAX_STATUS_LABEL_LEN + 1))
     msg_id = Column('msg_id', String(constants.MAX_MSG_ID_LEN + 1), ForeignKey('secure_message.msg_id'))
     actor = Column('actor', String(constants.MAX_STATUS_ACTOR_LEN + 1))
--- a/secure_message/repository/database.py
+++ b/secure_message/repository/database.py
@@ -24,7 +24,7 @@ class SecureMessage(db.Model):
     msg_id = Column("msg_id", String(constants.MAX_MSG_ID_LEN), unique=True, index=True)
     subject = Column("subject", String(constants.MAX_SUBJECT_LEN + 1))
     body = Column("body", String(constants.MAX_BODY_LEN + 1))
-    thread_id = Column("thread_id", String(constants.MAX_THREAD_LEN + 1), index=True)
+    thread_id = Column("thread_id", String(constants.MAX_THREAD_LEN + 1), ForeignKey('conversation.id'), index=True)
     collection_case = Column("collection_case", String(constants.MAX_COLLECTION_CASE_LEN + 1))
     ru_id = Column("ru_id", String(constants.MAX_RU_ID_LEN + 1))
     collection_exercise = Column("collection_exercise", String(constants.MAX_COLLECTION_EXERCISE_LEN + 1))
@@ -169,3 +169,20 @@ class Events(db.Model):
         self.msg_id = msg_id
         self.event = event
         self.date_time = datetime.now(timezone.utc)
+
+
+class Conversation(db.Model):
+    """Conversation table model"""
+    __tablename__ = "conversation"
+
+    id = Column('id', String(length=60), primary_key=True)
+    is_closed = Column('is_closed', Boolean())
+    closed_by = Column('closed_by', String())
+    closed_by_uuid = Column('closed_by_uuid', String(length=60))
+    closed_at = Column('closed_at', DateTime())
+
+    def __init__(self, is_closed=False, closed_by='', closed_by_uuid='', closed_at=None):
+        self.is_closed = is_closed
+        self.closed_by = closed_by
+        self.closed_by_uuid = closed_by_uuid
+        self.closed_at = closed_at
--- a/secure_message/repository/database.py
+++ b/secure_message/repository/database.py
@@ -137,7 +137,7 @@ class Status(db.Model):
     """Label Assignment table model"""
     __tablename__ = "status"
 
-    id = Column('id', Integer(), primary_key=True, index=True)
+    id = Column('id', Integer(), primary_key=True)
     label = Column('label', String(constants.MAX_STATUS_LABEL_LEN + 1))
     msg_id = Column('msg_id', String(constants.MAX_MSG_ID_LEN + 1), ForeignKey('secure_message.msg_id'))
     actor = Column('actor', String(constants.MAX_STATUS_ACTOR_LEN + 1))
@@ -183,7 +183,7 @@ class Conversation(db.Model):
     """Conversation table model"""
     __tablename__ = "conversation"
 
-    id = Column('id', String(length=60), primary_key=True)
+    id = Column('id', String(length=60), primary_key=True, index=True)
     is_closed = Column('is_closed', Boolean())
     closed_by = Column('closed_by', String())
     closed_by_uuid = Column('closed_by_uuid', String(length=60))
--- a/secure_message/repository/database.py
+++ b/secure_message/repository/database.py
@@ -30,6 +30,8 @@ class SecureMessage(db.Model):
     collection_exercise = Column("collection_exercise", String(constants.MAX_COLLECTION_EXERCISE_LEN + 1))
     survey = Column("survey", String(constants.MAX_SURVEY_LEN + 1))
     from_internal = Column('from_internal', Boolean())
+    sent_at = Column('sent_at', DateTime(), default=datetime.utcnow)
+    read_at = Column('read_at', DateTime())
 
     statuses = relationship('Status', backref='secure_message', lazy="dynamic")
     events = relationship('Events', backref='secure_message', order_by='Events.date_time', lazy="dynamic")
@@ -37,7 +39,7 @@ class SecureMessage(db.Model):
     __table_args__ = (Index("idx_ru_survey_cc", "ru_id", "survey", "collection_case", "collection_exercise"), )
 
     def __init__(self, msg_id="", subject="", body="", thread_id="", collection_case='',
-                 ru_id='', survey='', collection_exercise='', from_internal=False):
+                 ru_id='', survey='', collection_exercise='', from_internal=False, read_at=None):
 
         logger.debug(f"Initialised Secure Message entity: msg_id: {id}")
         self.msg_id = msg_id
@@ -49,6 +51,12 @@ class SecureMessage(db.Model):
         self.survey = survey
         self.collection_exercise = collection_exercise
         self.from_internal = from_internal
+        self.read_at = read_at
+
+    def __repr__(self):
+        return f'<SecureMessage(msg_id={self.msg_id} subject={self.subject} body={self.body} thread_id={self.thread_id}' \
+               f' collection_case={self.collection_case} ru_id={self.ru_id} collection_exercise={self.collection_exercise}' \
+               f' survey={self.survey} from_internal={self.from_internal} sent_at={self.sent_at} read_at={self.read_at})>'
 
     def set_from_domain_model(self, domain_model):
         """set dbMessage attributes to domain_model attributes"""
--- a/secure_message/repository/database.py
+++ b/secure_message/repository/database.py
@@ -189,7 +189,7 @@ class Conversation(db.Model):
     closed_by_uuid = Column('closed_by_uuid', String(length=60))
     closed_at = Column('closed_at', DateTime())
 
-    def __init__(self, is_closed=False, closed_by='', closed_by_uuid='', closed_at=None):
+    def __init__(self, is_closed=False, closed_by=None, closed_by_uuid=None, closed_at=None):
         self.is_closed = is_closed
         self.closed_by = closed_by
         self.closed_by_uuid = closed_by_uuid
--- a/secure_message/repository/modifier.py
+++ b/secure_message/repository/modifier.py
@@ -131,8 +131,8 @@ class Modifier:
             bound_logger.info("Re-opening conversation")
             metadata.is_closed = False
             metadata.closed_at = None
-            metadata.closed_by = ''
-            metadata.closed_by_uuid = ''
+            metadata.closed_by = None
+            metadata.closed_by_uuid = None
             db.session.commit()
         except SQLAlchemyError:
             db.session.rollback()
--- a/secure_message/repository/modifier.py
+++ b/secure_message/repository/modifier.py
@@ -110,7 +110,7 @@ class Modifier:
         bound_logger.unbind('conversation_id', 'user_id')
 
     @staticmethod
-    def open_conversation(metadata, user, session=db.session):
+    def open_conversation(metadata, user):
         bound_logger = logger.bind(conversation_id=metadata.id, user_id=user.user_uuid)
 
         try:
@@ -119,9 +119,9 @@ class Modifier:
             metadata.closed_at = None
             metadata.closed_by = ''
             metadata.closed_by_uuid = ''
-            session.commit()
+            db.session.commit()
         except SQLAlchemyError:
-            session.rollback()
+            db.session.rollback()
             logger.exception("Error saving metadata to conversation")
 
         bound_logger.info("Sucessfully re-opened conversation")
--- a/secure_message/repository/modifier.py
+++ b/secure_message/repository/modifier.py
@@ -137,5 +137,5 @@ class Modifier:
             db.session.rollback()
             logger.exception("Error saving metadata to conversation")
 
-        bound_logger.info("Sucessfully re-opened conversation")
+        bound_logger.info("Successfully re-opened conversation")
         bound_logger.unbind('conversation_id', 'user_id')
--- a/secure_message/repository/modifier.py
+++ b/secure_message/repository/modifier.py
@@ -117,7 +117,8 @@ class Modifier:
             db.session.commit()
         except SQLAlchemyError:
             db.session.rollback()
-            logger.exception("Error saving metadata")
+            logger.exception("Database error occured while closing conversation")
+            raise InternalServerError(description="Database error occured while closing conversation")
 
         bound_logger.info("Successfully closed conversation")
         bound_logger.unbind('conversation_id', 'user_id')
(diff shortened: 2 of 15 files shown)


secure_message/repository/database.py
```diff
@@ -169,3 +169,20 @@ def __init__(self, msg_id='', event=''):
         self.msg_id = msg_id
         self.event = event
         self.date_time = datetime.now(timezone.utc)
+
+
+class Conversation(db.Model):
+    """Conversation table model"""
+    __tablename__ = "conversation"
+
+    id = Column('id', String(length=60), primary_key=True)
```
MIslam92: Random Q how come in every other table we set the id to a Integer but here its set to a String?

secure_message/repository/modifier.py
```diff
@@ -84,3 +87,42 @@ def del_unread(message, user):
             Saver().save_msg_event(message['msg_id'], EventsApi.READ.value)
         Modifier.remove_label(unread, message, user)
         return True
+
+    @staticmethod
+    def close_conversation(metadata, user):
+        bound_logger = logger.bind(conversation_id=metadata.id, user_id=user.user_uuid)
+        bound_logger.info("Getting user details")
+        user_details = InternalUserService.get_user_details(user.user_uuid)
+        bound_logger.info("Sucessfully retreived user details")
```
MIslam92: i before e except after c? i.e. retrieved spelling also 'Successfully' as well

secure_message/repository/modifier.py
```diff
@@ -84,3 +87,42 @@ def del_unread(message, user):
...
+            metadata.is_closed = True
+            metadata.closed_at = datetime.now(timezone.utc)
+            metadata.closed_by = f"{user_details.get('firstName')} {user_details.get('lastName')}"
+            metadata.closed_by_uuid = user.user_uuid
+            db.session.commit()
+        except SQLAlchemyError:
+            db.session.rollback()
+            logger.exception("Error saving metadata")
+
+        bound_logger.info("Sucessfully closed conversation")
```
MIslam92: Spelling here again

secure_message/repository/modifier.py
```diff
@@ -84,3 +87,42 @@ def del_unread(message, user):
...
+            metadata.is_closed = False
+            metadata.closed_at = None
+            metadata.closed_by = ''
+            metadata.closed_by_uuid = ''
+            db.session.commit()
+        except SQLAlchemyError:
+            db.session.rollback()
+            logger.exception("Error saving metadata to conversation")
+
+        bound_logger.info("Sucessfully re-opened conversation")
```
MIslam92: 'Successfully' again

secure_message/repository/retriever.py
```diff
@@ -242,6 +243,20 @@ def _retrieve_thread_for_internal_user(thread_id):
...
+        result = Conversation.query.filter(Conversation.id == thread_id)
+        try:
+            return result.one()
+        except NoResultFound:
+            logger.info("No conversation found", thread_id=thread_id)
+            raise NotFound(description=f"Conversation with thread_id '{thread_id}' does not exist")
+        except MultipleResultsFound:
+            # We should never hit this code, but if we somehow did then a 500 would be accurate
+            # as something would be seriously wrong in the database.
+            logger.error("Mulitple results found for conversation", thread_id=thread_id)
```
MIslam92: 'Multiple'

secure_message/resources/threads.py
```diff
@@ -20,14 +22,73 @@ def get(thread_id):
...
+        request_data = request.get_json()
+        metadata = Retriever.retrieve_conversation_metadata(thread_id)
+        ThreadById._validate_request(request_data, metadata)
+
+        if 'is_closed' in request_data:
+            if request_data.get('is_closed'):
+                bound_logger.info("About to close conversation")
+                Modifier.close_conversation(metadata, g.user)
+            else:
+                bound_logger.info("About re-open conversation")
```
MIslam92: "About to re-open conversation"?

secure_message/repository/modifier.py
insacuri: Fixed

secure_message/repository/retriever.py
insacuri: Fixed

secure_message/resources/threads.py
insacuri: Fixed

secure_message/resources/threads.py
```diff
@@ -20,14 +22,73 @@ def get(thread_id):
...
+
+        bound_logger.info("Attempting to modify metadata for thread")
+        request_data = request.get_json()
+        metadata = Retriever.retrieve_conversation_metadata(thread_id)
+        ThreadById._validate_request(request_data, metadata)
+
+        if 'is_closed' in request_data:
+            if request_data.get('is_closed'):
+                bound_logger.info("About to close conversation")
+                Modifier.close_conversation(metadata, g.user)
```
ajmaddaford: What should happen if this fails? We shouldn't return a 204.

secure_message/resources/threads.py
```diff
@@ -20,14 +22,73 @@ def get(thread_id):
...
+            else:
+                bound_logger.info("About re-open conversation")
+                Modifier.open_conversation(metadata, g.user)
```
ajmaddaford: Same comment for closing, what should happen if this fails?

secure_message/repository/database.py
insacuri: I'll mention it here incase anyone else asks.  This table will, at some point, represent the conversation.  Theres no point in calling the id of it 'conversation_id' because we're in the conversation table already so we know what the ID is for already.  So we'll just call it id

secure_message/repository/retriever.py
```diff
@@ -242,6 +243,20 @@ def _retrieve_thread_for_internal_user(thread_id):
...
 
+    @staticmethod
+    def retrieve_conversation_metadata(thread_id):
+        result = Conversation.query.filter(Conversation.id == thread_id)
+        try:
+            return result.one()
+        except NoResultFound:
+            logger.info("No conversation found", thread_id=thread_id)
+            raise NotFound(description=f"Conversation with thread_id '{thread_id}' does not exist")
+        except MultipleResultsFound:
```
ajmaddaford: I don't see how this would _ever_ happen if we're querying on the primary key. If we can't write a test to test this, we shouldn't have it in the code.

secure_message/resources/messages.py
```diff
@@ -34,11 +34,16 @@ def post(self):
 
         if 'msg_id' in post_data:
             raise BadRequest(description="Message can not include msg_id")
+        if post_data.get('thread_id') in post_data:
```
ajmaddaford: You can drop the `in post_data`.

`post_data.get('thread_id')` will return `None` or a `<thread_id>`, so this statement is actually looking for a key that matches `<thread_id>` in post_data.

secure_message/resources/threads.py
```diff
@@ -20,14 +22,73 @@ def get(thread_id):
...
             messages.append(msg)
+
+        if conversation_metadata.is_closed:
+            return jsonify({"messages": add_users_and_business_details(messages),
+                            "is_closed": conversation_metadata.is_closed,
+                            "closed_by": conversation_metadata.closed_by,
+                            "closed_by_uuid": conversation_metadata.closed_by_uuid,
+                            "closed_at": conversation_metadata.closed_at})
+
         return jsonify({"messages": add_users_and_business_details(messages)})
```
ajmaddaford: We should consider returning `is_closed: false`, as this could simplify a clients use of the api - `is_closed` would always be present removing the need to check for it's existence before getting the value.

secure_message/resources/threads.py
```diff
@@ -20,14 +22,73 @@ def get(thread_id):
         logger.info("Getting messages from thread", thread_id=thread_id, user_uuid=g.user.user_uuid)
 
         conversation = Retriever().retrieve_thread(thread_id, g.user)
+        conversation_metadata = Retriever.retrieve_conversation_metadata(thread_id)
```
ajmaddaford: Does this return a 404 if the conversation metadata doesn't exist? (it won't exist for all current conversations until they're closed)

tests/app/test_modifier.py
```diff
@@ -22,11 +23,16 @@ class ModifyTestCaseHelper:
 
     def populate_database(self, record_count=0, mark_as_read=True):
         """Adds a sppecified number of Messages to the db"""
+        msg_id = str(uuid.uuid4())
```
ajmaddaford: Why is `msg_id` being instantiated here? (it's also done on line 29).

secure_message/repository/retriever.py
insacuri: It wouldn't happen.  I only put in there for completeness as .one() can throw two different exceptions.  If it's valid to not handle an exception that can't be thrown then that's fine by me as it's less code

secure_message/repository/database.py
```diff
@@ -177,3 +177,20 @@ def __init__(self, msg_id='', event=''):
...
+    is_closed = Column('is_closed', Boolean())
+    closed_by = Column('closed_by', String())
+    closed_by_uuid = Column('closed_by_uuid', String(length=60))
+    closed_at = Column('closed_at', DateTime())
+
+    def __init__(self, is_closed=False, closed_by='', closed_by_uuid='', closed_at=None):
```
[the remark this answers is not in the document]
insacuri: I was following the other examples where string values defaulted to empty strings instead of nothing.  If that's incorrect then I can change it

secure_message/resources/threads.py
```diff
@@ -20,14 +22,73 @@ def get(thread_id):
...
+            bound_logger.info('Request must set accept content type "application/json" in header.')
+            raise BadRequest(description="'application/json' content type in header missing")
+
+        bound_logger.info("Attempting to modify metadata for thread")
+        request_data = request.get_json()
+        metadata = Retriever.retrieve_conversation_metadata(thread_id)
+        ThreadById._validate_request(request_data, metadata)
+
+        if 'is_closed' in request_data:
+            if request_data.get('is_closed'):
```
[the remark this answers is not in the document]
insacuri: Removed it

secure_message/resources/threads.py
insacuri: Now it does yea

secure_message/resources/threads.py
insacuri: Returns a 500 now

secure_message/resources/threads.py
insacuri: Returns a 500 now

tests/app/test_modifier.py
```diff
@@ -47,17 +52,37 @@ def populate_database(self, record_count=0, mark_as_read=True):
                             VALUES('{EventsApi.READ.value}', '{msg_id}', '2017-02-03 00:00:00')'''
                     con.execute(query)
 
+        return msg_id
```
MIslam92: Relocate the `msg_id = str(uuid.uuid4())` to outside the `for` to resolve the reference error?

insacuri: Fixed, but in a slightly different way

secure_message/repository/database.py
atrozzo: if the id is going to be used in the queries liek we do for the thread id of msg id we should create an index.

secure_message/repository/modifier.py
```diff
@@ -97,3 +99,45 @@ def mark_message_as_read(message, user):
...
+    @staticmethod
+    def close_conversation(metadata, user):
+        bound_logger = logger.bind(conversation_id=metadata.id, user_id=user.user_uuid)
+        bound_logger.info("Getting user details")
+
+        user_details = internal_user_service.get_user_details(user.user_uuid)
+        bound_logger.info("Successfully retrieved user details")
+
+        try:
+            bound_logger.info("Closing conversation")
```
atrozzo: what about add conversation id in the log file so at leats we have some trace of it in case ?

secure_message/repository/modifier.py
```diff
@@ -97,3 +99,45 @@ def mark_message_as_read(message, user):
...
+            metadata.is_closed = True
+            metadata.closed_at = datetime.utcnow()
+            metadata.closed_by = f"{user_details.get('firstName')} {user_details.get('lastName')}"
+            metadata.closed_by_uuid = user.user_uuid
+            db.session.commit()
+        except SQLAlchemyError:
+            db.session.rollback()
+            logger.exception("Database error occured while closing conversation")
```
atrozzo: the same here , you want to give more info that you can in the error.

secure_message/resources/messages.py
```diff
@@ -34,11 +34,18 @@ def post(self):
 
         if 'msg_id' in post_data:
             raise BadRequest(description="Message can not include msg_id")
+        if post_data.get('thread_id'):
+            conversation_metadata = Retriever.retrieve_conversation_metadata(post_data.get('thread_id'))
+            if conversation_metadata is None:
+                abort(404)
```
atrozzo: I think this shuld be more aboort(404, "you message about why di dnot work")

secure_message/repository/modifier.py
insacuri: It is, that's what the bound logger is doing.

secure_message/resources/messages.py
insacuri: It's a 404, so what went wrong should be obvious.  From what I remember there is some default text that get's added when you return it, but I'll double check that now

secure_message/repository/database.py
insacuri: Index added

secure_message/repository/modifier.py
insacuri: You actually caught something I hadn't noticed!  There was a bound logger already there that I wasn't using.  Good catch ^^

secure_message/resources/messages.py
insacuri: I've removed this now.  I realised that there's going to be a time where there won't be a conversation record for every thread id.

secure_message/repository/modifier.py
```diff
@@ -97,3 +99,45 @@ def mark_message_as_read(message, user):
...
+    @staticmethod
+    def open_conversation(metadata, user):
+        bound_logger = logger.bind(conversation_id=metadata.id, user_id=user.user_uuid)
+
+        try:
+            bound_logger.info("Re-opening conversation")
+            metadata.is_closed = False
+            metadata.closed_at = None
+            metadata.closed_by = ''
+            metadata.closed_by_uuid = ''
```
ajmaddaford: `closed_by` and `closed_by_uuid` should be set to `None` to match what is done when a `Conversation` class is instantiated

secure_message/resources/threads.py
```diff
@@ -21,13 +22,77 @@ def get(thread_id):
         logger.info("Getting messages from thread", thread_id=thread_id, user_uuid=g.user.user_uuid)
 
         conversation = Retriever().retrieve_thread(thread_id, g.user)
+        conversation_metadata = Retriever.retrieve_conversation_metadata(thread_id)
+        if conversation_metadata is None:
+            abort(404)
```
ajmaddaford: This causes a http 500 error in response ops when viewing any conversations that don't have a conversation record. All old conversations (started prior to deploying this pull request) won't have a conversation record.

secure_message/resources/threads.py
```diff
@@ -21,13 +22,77 @@ def get(thread_id):
...
+
+    @staticmethod
+    def _validate_request(request_data, metadata):
+        """Used to validate data within request body"""
+        bound_logger = logger.bind(thread_id=metadata.id, user_uuid=g.user.user_uuid)
+        # Check if it's empty
+        if not request_data:
+            bound_logger.error('No properties provided')
+            raise BadRequest(description="No properties provided")
+        if not isinstance(request_data['is_closed'], bool):
```
ajmaddaford: If you send a request without the `is_closed` property e.g. `{ "fake_property": 12 }` a 500 error is generated. It would be better to catch the `KeyError` when accessing `request_data['is_closed']` and return a 400 error.

secure_message/resources/threads.py
insacuri: Fixed this and tested it for myself.

secure_message/resources/threads.py
insacuri: Fixed it

secure_message/resources/threads.py
```diff
@@ -21,13 +22,77 @@ def get(thread_id):
...
+            bound_logger.error('Invalid value provided')
+            raise BadRequest(description="Invalid label provided")
```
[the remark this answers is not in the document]
insacuri: Good spot, fixed it

secure_message/resources/threads.py
```diff
@@ -79,9 +79,14 @@ def _validate_request(request_data, metadata):
         if not request_data:
             bound_logger.error('No properties provided')
             raise BadRequest(description="No properties provided")
+        try:
+            request_data['is_closed']
+        except KeyError:
+            bound_logger.error('Invalid properties provided')
+            raise BadRequest(description="Only 'is_closed' property may be provided")
```
ajmaddaford: I think catching the KeyError would only make sense if it was combined with the key access in the type checking below. Seeing this makes me think it's clearer to do it this way though, but using an if:
```Python
if 'is_closed' not in request_data:
    bound_logger.error('Invalid properties provided')
    raise BadRequest(description="Only 'is_closed' property may be provided")
```

insacuri: Fixed

secure_message/resources/threads.py
```diff
@@ -23,16 +23,14 @@ def get(thread_id):
...
-            abort(404)
 
         logger.info("Successfully retrieved messages from thread", thread_id=thread_id, user_uuid=g.user.user_uuid)
         messages = []
         for message in conversation.all():
             msg = message.serialize(g.user, body_summary=False)
             messages.append(msg)
 
-        if conversation_metadata.is_closed:
+        if conversation_metadata is not None and conversation_metadata.is_closed:
```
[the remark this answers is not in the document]
insacuri: Fixed

secure_message/resources/threads.py
```diff
@@ -21,13 +22,78 @@ def get(thread_id):
...
             msg = message.serialize(g.user, body_summary=False)
             messages.append(msg)
-        return jsonify({"messages": add_users_and_business_details(messages)})
+
+        if conversation_metadata and conversation_metadata.is_closed:
+            return jsonify({"messages": add_users_and_business_details(messages),
+                            "is_closed": conversation_metadata.is_closed,
+                            "closed_by": conversation_metadata.closed_by,
+                            "closed_by_uuid": conversation_metadata.closed_by_uuid,
+                            "closed_at": conversation_metadata.closed_at})
```
ajmaddaford: This returns a date as a string e.g. `"closed_at": "Tue, 12 Jun 2018 09:48:36 GMT"`. This should be an ISO 8601 date time e.g. `2018-06-12T09:48:36Z`

insacuri: Done

PR #208, opened 2018-06-06
Batch respondent and business calls

ryangrundy7:

https://trello.com/c/cJMHZnbR/216-update-secure-message-to-use-new-party-service-batch-call-5

The purpose of this PR is to update secure messaging to use the batch call in the party service repo for the businesses and respondents.

**How to review**

- Run make test and all tests should pass

- Run the acceptance tests to get some messages and you should be able to see all messages on R-ops

### DIFF
--- a/secure_message/api_mocks/party_service_mock.py
+++ b/secure_message/api_mocks/party_service_mock.py
@@ -9,20 +9,14 @@ class PartyServiceMock:
 
     def get_business_details(self, ru):
         """Return mock business details"""
-        business_details = []
-        for ru_id in self._business_details:
-            if ru_id in ru:
-                business_details.append(self._business_details.get(ru_id))
+        business_details = [self._business_details.get(ru_id) for ru_id in self._business_details if ru_id in ru]
         if not business_details:
             logger.error(f"no details for ru:{ru} found in mock party service")
         return business_details
 
     def get_user_details(self, uuid):
         """Return mock user details"""
-        user_details = []
-        for uuids in self._respondent_ids:
-            if uuids in uuid:
-                user_details.append(self._respondent_ids)
+        user_details = [self._respondent_ids for uuids in self._respondent_ids if uuids in uuid]
         if not user_details:
             logger.error(f'Party service failed for uuid:{uuid}')
             user_details = None
--- a/secure_message/api_mocks/party_service_mock.py
+++ b/secure_message/api_mocks/party_service_mock.py
@@ -15,10 +15,13 @@ class PartyServiceMock:
         return business_details
 
     def get_user_details(self, uuid):
+        return self.get_users_details([uuid])
+
+    def get_users_details(self, uuids):
         """Return mock user details"""
-        user_details = [self._respondent_ids for uuids in self._respondent_ids if uuids in uuid]
+        user_details = [self._respondent_ids for uuid in self._respondent_ids if uuid in uuids]
         if not user_details:
-            logger.error(f'Party service failed for uuid:{uuid}')
+            logger.error(f'Party service failed for uuid:{uuids}')
             user_details = None
         return user_details
 
--- a/secure_message/api_mocks/party_service_mock.py
+++ b/secure_message/api_mocks/party_service_mock.py
@@ -19,7 +19,7 @@ class PartyServiceMock:
 
     def get_users_details(self, uuids):
         """Return mock user details"""
-        user_details = [self._respondent_ids for uuid in self._respondent_ids if uuid in uuids]
+        user_details = [self._respondent_ids[uuid] for uuid in self._respondent_ids if uuid in uuids]
         if not user_details:
             logger.error(f'Party service failed for uuid:{uuids}')
             user_details = None
--- a/secure_message/api_mocks/party_service_mock.py
+++ b/secure_message/api_mocks/party_service_mock.py
@@ -9,16 +9,24 @@ class PartyServiceMock:
 
     def get_business_details(self, ru):
         """Return mock business details"""
-        business_details = self._business_details.get(ru)
-        if business_details is None:
+        business_details = []
+        for ru_id in self._business_details:
+            if ru_id in ru:
+                business_details.append(self._business_details.get(ru_id))
+        if not business_details:
             logger.error(f"no details for ru:{ru} found in mock party service")
+            business_details = None
         return business_details
 
     def get_user_details(self, uuid):
         """Return mock user details"""
-        user_details = self._respondent_ids.get(uuid)
-        if user_details is None:
+        user_details = []
+        for uuids in self._respondent_ids:
+            if uuids in uuid:
+                user_details.append(self._respondent_ids)
+        if not user_details:
             logger.error(f'Party service failed for uuid:{uuid}')
+            user_details = None
         return user_details
 
     _business_details = {'c614e64e-d981-4eba-b016-d9822f09a4fb': {"id": "c614e64e-d981-4eba-b016-d9822f09a4fb",
--- a/secure_message/api_mocks/party_service_mock.py
+++ b/secure_message/api_mocks/party_service_mock.py
@@ -7,11 +7,11 @@ logger = wrap_logger(logging.getLogger(__name__))
 
 class PartyServiceMock:
 
-    def get_business_details(self, ru):
+    def get_business_details(self, ru_ids):
         """Return mock business details"""
-        business_details = [self._business_details.get(ru_id) for ru_id in self._business_details if ru_id in ru]
+        business_details = [self._business_details.get(ru_id) for ru_id in self._business_details if ru_id in ru_ids]
         if not business_details:
-            logger.error(f"no details for ru:{ru} found in mock party service")
+            logger.error("At least 1 ru_id missing in mock party service", ru_ids=ru_ids)
         return business_details
 
     def get_user_details(self, uuid):
--- a/secure_message/api_mocks/party_service_mock.py
+++ b/secure_message/api_mocks/party_service_mock.py
@@ -15,7 +15,6 @@ class PartyServiceMock:
                 business_details.append(self._business_details.get(ru_id))
         if not business_details:
             logger.error(f"no details for ru:{ru} found in mock party service")
-            business_details = None
         return business_details
 
     def get_user_details(self, uuid):
--- a/secure_message/common/utilities.py
+++ b/secure_message/common/utilities.py
@@ -81,7 +81,7 @@ def get_to_details(messages):
             message.update({"@msg_to": [internal_user_service.get_user_details(message["msg_to"][0])]})
 
     # Then resolve msg_to details for messages going from ONS user (internal) to respondant (external)
-    to_details = party.get_users_details(get_external_user_uuid_list_from_msg_to_key_in_internal_messages(messages))
+    to_details = party.get_users_details(get_external_user_uuid_list(messages))
     internal_messages = [message for message in messages if message['from_internal'] is True]
 
     for message in messages:
@@ -99,7 +99,7 @@ def get_from_details(messages):
             message.update({"@msg_from": internal_user_service.get_user_details(message["msg_from"])})
 
     # Then resolve msg_from details for messages going from respondant (external) to ONS user (internal)
-    from_details = party.get_users_details(get_external_user_uuid_list_from_msg_from_key_in_external_messages(messages))
+    from_details = party.get_users_details(get_external_user_uuid_list(messages))
     external_messages = [message for message in messages if message['from_internal'] is False]
 
     for message in messages:
@@ -109,25 +109,20 @@ def get_from_details(messages):
     return messages
 
 
-def get_external_user_uuid_list_from_msg_from_key_in_external_messages(messages):
+def get_external_user_uuid_list(messages):
     """Compiles a list of all unique the external user (respondant) uuids from a list of messages"""
-    uuid_from = []
+    external_user_uuids = []
 
     external_msgs = [message for message in messages if message['from_internal'] is False]
     for uuid in external_msgs:
-        if uuid["msg_from"] not in uuid_from:
-            uuid_from.append(uuid["msg_from"])
-    return uuid_from
+        if uuid["msg_from"] not in external_user_uuids:
+            external_user_uuids.append(uuid["msg_from"])
 
-
-def get_external_user_uuid_list_from_msg_to_key_in_internal_messages(messages):
-    """Creates a list of unique uuids for external (respondant) users from a list of messages"""
-    uuid_to = []
     internal_messages = [message for message in messages if message['from_internal'] is True]
     for uuid in internal_messages:
-        if uuid["msg_to"][0] not in uuid_to:
-            uuid_to.append(uuid["msg_to"][0])
-    return uuid_to
+        if uuid["msg_to"][0] not in external_user_uuids:
+            external_user_uuids.append(uuid["msg_to"][0])
+    return external_user_uuids
 
 
 def add_business_details(messages):
--- a/secure_message/common/utilities.py
+++ b/secure_message/common/utilities.py
@@ -131,8 +131,7 @@ def get_external_user_uuid_list(messages):
 
 
 def add_business_details(messages):
-    """Adds business details"""
-
+    """Adds a @ru_id key every message in a list of messages."""
     ru_ids = []
 
     for message in messages:
--- a/secure_message/common/utilities.py
+++ b/secure_message/common/utilities.py
@@ -116,27 +116,25 @@ def add_from_details(messages):
 
 def get_external_user_uuid_list(messages):
     """Compiles a list of all unique the external user (respondant) uuids from a list of messages"""
-    external_user_uuids = []
+    external_user_uuids = set()
 
     external_msgs = [message for message in messages if message['from_internal'] is False]
     for message in external_msgs:
-        if message["msg_from"] not in external_user_uuids:
-            external_user_uuids.append(message["msg_from"])
+        external_user_uuids.add(message["msg_from"])
 
     internal_messages = [message for message in messages if message['from_internal'] is True]
     for uuid in internal_messages:
-        if uuid["msg_to"][0] not in external_user_uuids:
-            external_user_uuids.append(uuid["msg_to"][0])
+        external_user_uuids.add(uuid["msg_to"][0])
+
     return external_user_uuids
 
 
 def add_business_details(messages):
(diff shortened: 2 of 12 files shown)


secure_message/api_mocks/party_service_mock.py
```diff
@@ -9,16 +9,23 @@ class PartyServiceMock:
 
     def get_business_details(self, ru):
         """Return mock business details"""
-        business_details = self._business_details.get(ru)
-        if business_details is None:
+        business_details = []
+        for ru_id in self._business_details:
+            if ru_id in ru:
+                business_details.append(self._business_details.get(ru_id))
```
dcdarrell9: This could be a list comp:
```
new_list = []
for i in old_list:
    if filter(i):
        new_list.append(expressions(i))
```

to:
`new_list = [expression(i) for i in old_list if filter(i)]`

secure_message/api_mocks/party_service_mock.py
```diff
@@ -9,16 +9,23 @@ class PartyServiceMock:
...
         return business_details
 
     def get_user_details(self, uuid):
         """Return mock user details"""
-        user_details = self._respondent_ids.get(uuid)
-        if user_details is None:
+        user_details = []
+        for uuids in self._respondent_ids:
+            if uuids in uuid:
+                user_details.append(self._respondent_ids)
```
dcdarrell9: Same with this, could be a list comp

secure_message/common/utilities.py
```diff
@@ -70,56 +70,76 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
-            user_details.append(detail)
+    for i in range(0, len_list):
+        if messages[i]['from_internal']:
+            msgs.append(messages[i])
 
-    return user_details
+    for uuid in msgs:
+        if uuid["msg_to"][0] not in uuid_to:
+            uuid_to.append(uuid["msg_to"][0])
+    return uuid_to
```
dcdarrell9: Again could be a list comp, but maybe readability comes in here

secure_message/api_mocks/party_service_mock.py
ryangrundy7: Done

secure_message/api_mocks/party_service_mock.py
ryangrundy7: Done

secure_message/common/utilities.py
ryangrundy7: Don't think i should change it because this if statement stops duplicating values into the list and i don't think i'd be able to do that in a list comp

dcdarrell9: I would have thought it would still be possible, as it gives the same output, but no problem 👍

secure_message/services/party_service.py
```diff
@@ -8,42 +10,41 @@
...
-                logger.debug(f"Party data failed for ru:{ru}", status_code=party_data.status_code, text=party_data.text,
-                             ru=ru)
-
-        return ru_dict
+        ru_list = []
+        payload = urlencode([("id", x) for x in rus])
+
+        party_data = requests.get(f"{current_app.config['RAS_PARTY_SERVICE']}party-api/v1/businesses",
+                                  auth=current_app.config['BASIC_AUTH'], verify=False, params=payload)
+        if party_data.status_code == 200:
```
dcdarrell9: Maybe we could put this in a try/except, but this looks fine 👍

secure_message/common/utilities.py
```diff
@@ -70,56 +70,76 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
     return messages, links
 
 
-def get_business_details_by_ru(rus):
-    """Function to retrieve business details from ru using the party service"""
+def get_from_details(messages):
+    """looks up the details for the from users"""
+    update_internal_messages_from(messages)
+    from_details = party.get_user_details(get_messages_from_external(messages))
+    external_messages = [x for x in messages if x['from_internal'] is False]
```
MIslam92: 'x' doesn't seem like a very explicit name maybe rename this to something more meaningful and makes sense in context

secure_message/common/utilities.py
```diff
@@ -70,56 +70,76 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
 
-def _get_to_details(message):
-    """looks up the details of all the to users"""
-    user_details = []
-    if message['from_internal']:
-        to_user_details_finder = party.get_user_details
-    else:
-        to_user_details_finder = internal_user_service.get_user_details
+    to_details = party.get_user_details(get_internal_messages_to_uuid(messages))
+    internal_messages = [x for x in messages if x['from_internal'] is True]
```
MIslam92: Same as previous comment

secure_message/common/utilities.py
```diff
@@ -70,56 +70,76 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
+        if msg in internal_messages:
+            msg.update({'@msg_to': to_details})
+
+    return messages
+
+
+def update_external_messages_to(messages):
+    for message in messages:
+        if not message["from_internal"]:
+            message.update({"@msg_to": internal_user_service.get_user_details(message["msg_to"][0])})
```
insacuri: Not sure if I'm misunderstanding this.  But you get all the external messages, then you repeatedly call the internal_user_service for each one?

secure_message/services/party_service.py
```diff
@@ -8,42 +10,41 @@
...
-
-        user_dict = self._users_cache.get(uuid)
-
-        if user_dict is None:
-            logger.info(f"Party Service: retrieve party user data for {uuid}")
-            party_data = requests.get(f"{current_app.config['RAS_PARTY_SERVICE']}party-api/v1/respondents/id/{uuid}",
-                                      auth=current_app.config['BASIC_AUTH'], verify=False)
+        """Retrieves the user details from the party service"""
+        user_list = None
+        if self._users_cache:
```
insacuri: I'm not sure this'll do what you think it does.  I think it'll cache it for THIS app, but no others.  When you've got multiple instances running then it'll hit this cache every so often.
Ideally this should be stored in redis

secure_message/services/party_service.py
```diff
@@ -8,42 +10,41 @@
...
+                                  auth=current_app.config['BASIC_AUTH'], verify=False, params=payload)
+        if party_data.status_code == 200:
+            ru_list = party_data.json()
+            logger.debug(f"Party data retrieved for ru:{rus}")
+        else:
+            logger.debug(f"Party data failed for ru:{rus}", status_code=party_data.status_code, text=party_data.text,
+                         ru=rus)
+        return ru_list
 
     def get_user_details(self, uuid):
```
insacuri: Is this one or many uuids?  If so, it should reflect the plural nature of it

secure_message/validation/user.py
```diff
@@ -22,4 +22,5 @@ def is_valid_internal_user(uuid):
 
     @staticmethod
     def is_valid_respondent(uuid):
-        return party.get_user_details(uuid) is not None
+        uuid_list = [uuid]
+        return party.get_user_details(uuid_list) is not None
```
insacuri: Might as well inline this ```return party.get_user_details([uuid]) is not None```

secure_message/services/party_service.py
insacuri: Done, the whole thing isn't in a try/catch but the raise_for_status is

secure_message/common/utilities.py
atrozzo: I disagree @MIslam92 . It s a pythonic comprehension to do it . Are you refereing to the function name ?

insacuri: So much has changed, everyone reviewing this should ignore all the comments that came before it

secure_message/common/utilities.py
```diff
@@ -70,68 +70,75 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
+    aren't guaranteed to be provided so we're not breaking the contract by doing this.
+    """
+    external_user_details = party.get_users_details(get_external_user_uuid_list(messages))
+    for message in messages:
+        if message["from_internal"]:
+            message.update({"@msg_from": internal_user_service.get_user_details(message["msg_from"])})
+        else:
+            for detail in external_user_details:
+                if detail['id'] == message['msg_from']:
+                    message.update({'@msg_from': detail})
```
ajmaddaford: If we took the data returned from the party service and structured it like:
```
details = {
    'party_uuid1': {
		...party_detail...
    },
    'party_uuid2': {
		...party_detail...
    },
    ...
}
```
we wouldn't need to loop over `external_user_details` for every message, we could do:
```
if details.get(msg_from):
	message.update({'@msg_from': details.get(msg_from)})
```

secure_message/services/party_service.py
```diff
@@ -1,49 +1,53 @@
...
+        uuid into a list of one"""
+        return self._get_user_details_from_party_service([uuid])
+
+    def get_users_details(self, uuids):
+        """Retrieves the user details from the party service"""
+        return self._get_user_details_from_party_service(uuids)
+
+    @staticmethod
+    def _get_user_details_from_party_service(uuids):
+        payload = urlencode([("id", uuid) for uuid in uuids])
```
ajmaddaford: I think `payload` is misleading, it makes me think of something we'd be posting, maybe `id_params`?

secure_message/services/party_service.py
```diff
@@ -1,49 +1,53 @@
...
-            if party_data.status_code == 200:
-                user_dict = party_data.json()
-                self._users_cache[uuid] = user_dict
-                logger.debug(f"Party data retrieved for uuid:{uuid}")
-            else:
-                logger.error(f'Party service failed for uuid:{uuid}', status_code=party_data.status_code, text=party_data.text,
-                             uuid=uuid)
-        return user_dict
+        """Retrieves the user details from the party service after converting single
+        uuid into a list of one"""
```
ajmaddaford: I think this comment is unnecessary, it's clearly creating a list of one uuid.

secure_message/common/utilities.py
```diff
@@ -70,68 +70,75 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
-
-    ru_list = []
+    """Adds a @ru_id key every message in a list of messages."""
+    ru_ids = []
 
     for message in messages:
-        if message['ru_id'] not in ru_list:
-            ru_list.append(message['ru_id'])
+        if message['ru_id'] not in ru_ids:
+            ru_ids.append(message['ru_id'])
```
ajmaddaford: Same comment as previously about using a set to avoid the need for the `if`.

secure_message/common/utilities.py
```diff
@@ -140,8 +147,9 @@ def add_business_details(messages):
 
 def add_users_and_business_details(messages):
     """Add both user and business details to messages based on data from party service"""
-    messages = add_to_and_from_details(messages)
-    logger.info("Sucessfully added to and from details")
+    messages = get_to_details(messages)
+    messages = get_from_details(messages)
```
ajmaddaford: I think these methods would express their intent more clearly by using `add_` rather than `get_` as they add the details to the messages passed to them, rather than extract the 'to' and 'from' details.

insacuri: Agreed, changed them all

secure_message/services/party_service.py
insacuri: Fixed

secure_message/common/utilities.py
insacuri: Updated with cleaner way

secure_message/common/utilities.py
insacuri: Done

secure_message/common/utilities.py
```diff
@@ -70,68 +70,75 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
-            user_details.append(detail)
+    external_msgs = [message for message in messages if message['from_internal'] is False]
+    for uuid in external_msgs:
+        if uuid["msg_from"] not in external_user_uuids:
+            external_user_uuids.append(uuid["msg_from"])
 
-    return user_details
+    internal_messages = [message for message in messages if message['from_internal'] is True]
+    for uuid in internal_messages:
+        if uuid["msg_to"][0] not in external_user_uuids:
```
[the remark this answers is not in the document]
insacuri: Done

secure_message/common/utilities.py
```diff
@@ -70,68 +70,73 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
-    """looks up the details of all the to users"""
-    user_details = []
-    if message['from_internal']:
-        to_user_details_finder = party.get_user_details
-    else:
-        to_user_details_finder = internal_user_service.get_user_details
 
-    for uuid in message['msg_to']:
+def get_external_user_uuid_list(messages):
+    """Compiles a list of all unique the external user (respondant) uuids from a list of messages"""
```
MIslam92: Should be` respondent`

secure_message/services/party_service.py
```diff
@@ -1,49 +1,52 @@
...
+        response = requests.get(f"{current_app.config['RAS_PARTY_SERVICE']}party-api/v1/businesses",
+                                auth=current_app.config['BASIC_AUTH'], verify=False, params=params)
+        try:
+            response.raise_for_status()
+        except HTTPError:
+            logger.debug("Business detail retrieval failed", status_code=response.status_code, text=response.text,
+                         ru_ids=ru_ids)
+            return []
+
+        logger.debug("Business details sucessfully retrieved", ru_ids=ru_ids)
```
MIslam92: `successfully`

tests/app/test_retriever.py
```diff
@@ -13,15 +13,15 @@
...
-from tests.app.test_utilities import BRES_SURVEY, get_args
+from tests.app.test_utilities import get_args
 
 
 class RetrieverTestCaseHelper:
 
     default_internal_actor = 'internal_actor'
     second_internal_actor = 'second_internal_actor'
     default_external_actor = 'external_actor'
+    BRES_SURVEY = "33333333-22222-3333-4444-88dc018a1a4c"
```
MIslam92: Should we have a `setUp` function here prior to remove the unresolved ref to all the engine refs in this class similar to how it is done in the `test_modifier.py` class?

insacuri: Doing this would require a complete restructuring of the tests.  Considering how long this PR has gone on for and how this PR actually tidied this bit up from what it was, it should be done in some other PR.

tests/app/test_party_service.py
```diff
@@ -22,83 +22,106 @@ def setUp(self):
...
         """Test get business details sends a request and returns data"""
-        ru = "1234"
+        ru_ids = ["c614e64e-d981-4eba-b016-d9822f09a4fb", "c614e64e-d981-4eba-b016-d9822f09a4f2"]
+        business_data_url = f"{self.app.config['RAS_PARTY_SERVICE']}party-api/v1/businesses?id={ru_ids[0]}&id={ru_ids[1]}"
+        mock_request.get(business_data_url, status_code=200, text='{"something": "else"}')
+        with self.app.app_context():
+            results = PartyService().get_business_details(ru_ids)
 
-        sut = PartyService()
+        self.assertEqual(results, {"something": "else"})
```
ajmaddaford: This isn't testing what it implies it is - if I added a uuid or removed one it would continue to pass.

tests/app/test_party_service.py
```diff
@@ -22,83 +22,106 @@ def setUp(self):
...
+
+    @requests_mock.mock()
+    def test_results_returned_from_get_business_get_executed_twice_for_different_ru(self, mock_request):
```
ajmaddaford: What is this testing?

tests/app/test_party_service.py
```diff
@@ -22,83 +22,106 @@ def setUp(self):
...
 
         self.assertTrue(mock_request.call_count == 2)
 
     @requests_mock.mock()
     def test_results_returned_from_get_business_details_returned_as_expected(self, mock_request):
         """Test get business details sends a request and returns data"""
-        ru = "1234"
-        business_data_url = f"{self.app.config['RAS_PARTY_SERVICE']}party-api/v1/businesses/id/{ru}"
+        ru_ids = ["b08c07c3-df28-4283-bb4c-c048729ce372"]
+        business_data_url = f"{self.app.config['RAS_PARTY_SERVICE']}party-api/v1/businesses?id={ru_ids[0]}"
```
ajmaddaford: Similar to previous comments, it would be clearer if we just checked the mock was called rather that checking for `text` that we've mocked.

tests/app/test_party_service.py
```diff
@@ -22,83 +22,106 @@ def setUp(self):
...
+    def test_multiple_business_uuid_one_invalid(self, mock_request):
+        """Test get business details sends a request and returns data"""
+        ru_ids = ["c614e64e-d981-4eba-b016-d9822f09a4fb", "12345"]
+        texts = "error"
+        business_data_url = f"{self.app.config['RAS_PARTY_SERVICE']}party-api/v1/businesses?id={ru_ids[0]}&id={ru_ids[1]}"
+        mock_request.get(business_data_url, status_code=400, reason="Invalid uuid", text=texts)
+        with self.app.app_context():
+            results = PartyService().get_business_details(ru_ids)
+
+        self.assertEqual(results, [])
```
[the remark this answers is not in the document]
insacuri: I don't understand.  Are you wanting the list of one to just be a non-uuid just to show that a non-uuid makes it fall over?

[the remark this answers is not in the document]
ajmaddaford: There is no specific code in secure message for dealing with an invalid uuid, so implying that's what we're testing is misleading. The only code we can test is what should happen if the service we're calling errors. We should create the minimal test for this, which is what should we expect from `get_business_details` when the party service returns a 400.

tests/app/test_mock_party.py
```diff
@@ -34,11 +33,11 @@ def test_get_business_details_multiple_ru(self):
...
                    ]
 
-        business_details = get_business_details_by_ru(list_ru)
+        business_details = party.get_business_details(list_ru)
 
-        self.assertEqual(business_details[0]['id'], list_ru[0])
+        self.assertEqual(business_details[0]['id'], list_ru[2])
         self.assertEqual(business_details[1]['id'], list_ru[1])
-        self.assertEqual(business_details[2]['id'], list_ru[2])
+        self.assertEqual(business_details[2]['id'], list_ru[0])
```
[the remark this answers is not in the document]
insacuri: Fixed

PR #211, opened 2018-06-11
Copy sent events into messages

insacuri:

https://trello.com/c/Fv73S3XV/223-migrate-events-data-into-securemessage-3

This is the second part of the migration away from the events table.  We need to have every message in the secure-message table to have the data stored in the events table, stored in itself.  This will copy every sent datetime from the events table into the sent_at field in secure_message.

**How to review**

- Create a number of conversations locally.  
- Open some of them to get read records.  
- Go into the database and delete any number of sent_at and read_at records in the secure-message table
- Run the script
- Check the database, all the sent_at and read_at records that were deleted should have reappeared and no others should have appeared.
- For piece of mind, maybe try it again, but delete a different number of records.


copy-sent-datetime-from-events.py
```diff
@@ -0,0 +1,33 @@
...
+from sqlalchemy.orm import sessionmaker
+from secure_message.repository.database import SecureMessage, Events
+from config import Config # NOQA
+
+engine = create_engine(Config.SQLALCHEMY_DATABASE_URI)
+Session = sessionmaker(bind=engine)
+session = Session()
+
+try:
+    messages_without_sent_at = session.query(SecureMessage).filter(SecureMessage.sent_at == None).all() # noqa
```
atrozzo: There is  any idea of what could be or will be in a the future the size of this result ? query.all could get maybe dangerous in case of many data. There is any need of maybe "yeld_per" need ?

insacuri: That's a fair point, it's easy to forget the number of rows I might be dealing with...
I'll find out how many, and if it's a lot (and I suspect it is).  I'll add the yield_per

copy-sent-datetime-from-events.py
```diff
@@ -0,0 +1,33 @@
...
+            message.sent_at = sent_event.date_time
+            session.add(message)
+    session.commit()
+except Exception as e:
+    session.rollback()
+    print(e)
+
+
+try:
+    messages_without_read_at = session.query(SecureMessage).filter(SecureMessage.read_at == None).all() # noqa
```
MIslam92: Same as above comment

copy-sent-datetime-from-events.py
MIslam92: Should this be `(SecureMessage.read_at is None)` as it is complaining about an comparison error

insacuri: Have you tried it yourself?  I think I tried doing it the way they wanted it and I couldn't get it to work.  I think there are other examples of that in the application code for the same reason

MIslam92: Yeah I tried this and it doesn't work so I'm happy with whats been decided atm after having a chat with Darrell.

MIslam92: Yeah I've tried this with the whole `.filter(SecureMessage.sent_at is None)` and it did not work. I'm happy with the current solution after having a chat with Darrell as well on some clarification.

PR #212, opened 2018-06-11
Add filter for closed conversations

dcdarrell9:
# What is the context of this PR?
- Changed the functionality of the query in retrieve_thread_list to be able to pass through an is_closed param which allows us to filter by open/closed
- Branched off [close conversation branch](https://github.com/ONSdigital/ras-secure-message/pull/207)

# How to review
- Tests all pass.
- Run alongside the branches in secure message and frontstage and send messages between the two to be able to test how everything works.
- Try different ways of sending/receiving messages and different amounts.

# Links
- Linked to [this branch in response-operations](https://github.com/ONSdigital/response-operations-ui/pull/167)
- Relevant[ changes in ras-frontstage](https://github.com/ONSdigital/ras-frontstage/pull/352)


tests/app/test_retriever.py
```diff
@@ -36,11 +36,11 @@ def add_secure_message(self, msg_id, subject="test", body="test", thread_id="Thr
                     '{thread_id}', '{collection_case}', '{ru_id}', '{survey}', '{collection_exercise}', '{from_internal}')'''
             con.execute(query)
 
-    def add_conversation(self, conversation_id):
+    def add_conversation(self, conversation_id, is_closed):
```
ajmaddaford: The api doesn't allow creating a conversation in a closed state, so I don't think it should be a parameter here i.e. the SQL should always set `is_closed` to `False`.

secure_message/repository/retriever.py
```diff
@@ -31,7 +31,7 @@ def unread_message_count(user):
         return result
 
     @staticmethod
-    def thread_count_by_survey(user, survey, label=None):
+    def thread_count_by_survey(survey, is_closed):
```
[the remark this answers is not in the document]
dcdarrell9: As we're doing a thread count by survey I think this is fine as it's only being used by internal, we don't differentiate between survey on external side yet(not sure if we will), its just one list of threads which we get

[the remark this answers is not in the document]
ajmaddaford: I think we should return a 403 error in that case to make it clear to anyone consuming the api or developing it.

secure_message/common/utilities.py
```diff
@@ -27,12 +27,15 @@ def get_options(args):
         if args.get(field):
             fields[field] = int(args.get(field))
 
-    if args.get('desc') and args.get('desc') == 'false':
+    if args.get('desc') == 'false':
```
shahi645: 🥇

secure_message/repository/retriever.py
```diff
@@ -31,40 +31,29 @@ def unread_message_count(user):
         return result
 
     @staticmethod
-    def thread_count_by_survey(user, survey, label=None):
+    def thread_count_by_survey(user, survey, is_closed):
         """Count users threads for a specific survey"""
 
+        if not user.is_internal:
+            logger.info("Thread count should be internal users only", user_uuid=g.user.user_uuid)
+            abort(403)
```
shahi645: I'd usually try to avoid using globals where possible, but not sure if this can be passed through. Just a thought

ajmaddaford: The checking of permission and abort should be done in the `ThreadCounter` resource. The retriever should just be concerned with getting data from the database and returning it.

secure_message/resources/threads.py
```diff
@@ -113,16 +114,14 @@ def get():
...
-            label = str(request.args.get('label'))
-            if label.lower() == 'unread':
-                return jsonify(name=label, total=Retriever.thread_count_by_survey(g.user, survey, label))
-            else:
-                logger.debug('Invalid label name', name=label, request=request.url)
-                raise BadRequest(description="Invalid label")
-        else:
-            return jsonify(total=Retriever.thread_count_by_survey(g.user, survey))
+        surveys = request.args.getlist('survey')
+        is_closed = bool(strtobool(request.args.get('is_closed')))
```
ajmaddaford: We should default this to 'false' if not provided as otherwise a call to `/v2/messages/count` without the `is_closed` param fails.

secure_message/resources/threads.py
```diff
@@ -118,7 +118,7 @@ class ThreadCounter(Resource):
     @staticmethod
     def get():
         surveys = request.args.getlist('survey')
-        is_closed = bool(strtobool(request.args.get('is_closed')))
+        is_closed = bool(strtobool(request.args.get('is_closed', False)))
```
ajmaddaford: This generates a `AttributeError: 'bool' object has no attribute 'lower'` error, `False` needs to be a string.

PR #216, opened 2018-06-25
Add conversation record for every conversation

MIslam92:
### What is the context of this PR?

This is the next step in the db migration and is needed to put the data to a consistent state and to allow for the following PR's to be mergeable:

https://github.com/ONSdigital/ras-secure-message/pull/212

https://github.com/ONSdigital/ras-frontstage/pull/352

Trello card: https://trello.com/c/wC4PUlNr/12-add-conversation-record-for-every-conversation-3

### How to review

Does it work? 

- Try removing an` id` from your `conversation` table and its corresponding key from the `secure_message `table 
- Check that when you run the script 
i.e. `python add-convo-record-for-every-convo.py` that it is added. 

- You can do a print statement to check the actual SQL that is ran and check what gets inserted into the `conversation` table.

### DIFF
--- a/add-convo-record-for-every-convo.py
+++ b/add-convo-record-for-every-convo.py
@@ -1,0 +1,24 @@
+from sqlite3 import DatabaseError
+import sys
+
+from sqlalchemy import create_engine, insert
+from sqlalchemy.orm import sessionmaker
+
+from config import Config
+from secure_message.repository.database import Conversation, SecureMessage
+
+engine = create_engine(Config.SQLALCHEMY_DATABASE_URI)
+Session = sessionmaker(bind=engine)
+db = Session()
+
+try:
+    subquery = db.query(Conversation.id)
+    distinct_ids = db.query(SecureMessage.thread_id).filter(~SecureMessage.thread_id.in_(subquery)).distinct()
+    for thread_id in distinct_ids:
+        db.execute(insert(Conversation).values({'id': thread_id, 'is_closed': 'False'}))
+        print('Added thread_id: {}'.format(thread_id[0]))
+    db.commit()
+except DatabaseError:
+    db.rollback()
+    print(DatabaseError)
+    sys.exit('Exiting')


add-convo-record-for-every-convo.py
```diff
@@ -0,0 +1,25 @@
...
+from sqlalchemy import create_engine, insert
+from sqlalchemy.orm import sessionmaker
+from secure_message.repository.database import SecureMessage, Conversation
+from config import Config
+
+engine = create_engine(Config.SQLALCHEMY_DATABASE_URI)
+Session = sessionmaker(bind=engine)
+db = Session()
+logging.basicConfig()
+logging.getLogger('sqlalchemy.engine').setLevel(logging.INFO)
```
dcdarrell9: The logs now aren't very useful, maybe add some context to say what's being done or remove?

add-convo-record-for-every-convo.py
```diff
@@ -0,0 +1,25 @@
+import logging
```
dcdarrell9: Alphabetised imports

add-convo-record-for-every-convo.py
```diff
@@ -0,0 +1,21 @@
...
+
+engine = create_engine(Config.SQLALCHEMY_DATABASE_URI)
+Session = sessionmaker(bind=engine)
+db = Session()
+
+try:
+    subquery = db.query(Conversation.id)
+    distinct_ids = db.query(SecureMessage.thread_id).filter(~SecureMessage.thread_id.in_(subquery)).distinct()
+    for thread_id in distinct_ids:
+        db.execute(insert(Conversation).values({"id": thread_id, "is_closed": "False"}))
```
ajmaddaford: I think it should at least log out the thread id so progress can be monitored, currently it sits at the command prompt until completion.

add-convo-record-for-every-convo.py
```diff
@@ -0,0 +1,21 @@
...
+    db.commit()
+except DatabaseError:
+    db.rollback()
+    print(DatabaseError)
```
ajmaddaford: We should exit the script with an error (see https://stackoverflow.com/questions/22633544/how-to-throw-error-and-exit-in-python). This means when it's run as a CloudFoundry task if it errors it will be visible in the task list as a failure (https://docs.cloudfoundry.org/devguide/using-tasks.html#list-tasks)

add-convo-record-for-every-convo.py
shahi645: i second this

shahi645: ```INFO:config:thread_id=('4f764286-13eb-4c7a-ab66-862e383afb03',) event='Adding the following threads:'```

PR #222, opened 2018-07-10
Fix bug to call Notify when a new message is sent from internal to ex…

gavinleeedwards:

We currently are not notifying respondents that they have received a secure message by email. The message is still being sent and received as expected, its the notification by email which is failing. The error in the logs is "User does not have an emailAddress specified" and was logged for every secure 

https://trello.com/c/p0IJOD4P

**What's changed**

An 'if' statement in the code the code was silently failing so the app thought that there was no email address for the respondent, even though there was one.  

**How to review**

All tests should still pass.
Send a secure message from internal to external and check that Notify gets called (this can only be done from preprod).

### DIFF
--- a/tests/app/test_app.py
+++ b/tests/app/test_app.py
@@ -262,7 +262,7 @@ class FlaskTestCase(unittest.TestCase):
                                                                        "status": "ACTIVE",
                                                                        "sampleUnitType": "BI"}, 200))
     @patch.object(AlertViaLogging, 'send')
-    def test_notify_called(self, mock_alerter, mock_party):
+    def test_notify_called(self, mock_alerter, _):
         """Test that Notify is called when sending a new secure message """
 
         url = "http://localhost:5050/message/send"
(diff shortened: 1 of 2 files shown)


tests/app/test_app.py
```diff
@@ -254,6 +254,21 @@ def test_case_service_called_on_sent_if_NotifyCaseService_is_set(self, case):
...
 
+    @patch.object(PartyServiceMock, 'get_user_details', return_value=({"id": "f62dfda8-73b0-4e0e-97cf-1b06327a6712",
+                                                                       "firstName": "Bhavana",
+                                                                       "emailAddress": "example@example.com",
+                                                                       "lastName": "Lincoln",
+                                                                       "telephone": "+443069990888",
+                                                                       "status": "ACTIVE",
+                                                                       "sampleUnitType": "BI"}, 200))
+    @patch.object(AlertViaLogging, 'send')
+    def test_notify_called(self, mock_alerter, mock_party):
```
gemmai95: I think you can change mock_party to _ for the codacy issue it is flaring.

PR #224, opened 2018-07-18
Stopped generating errors on empty thread list

ryangrundy7:
When a user requests an empty thread list on fronstage or R-Ops, a exception is generated on secure-message and party-service saying parameter id is required. This PR fixes it by checking if theres an empty list before making any calls to the party service and stops the generation of the exceptions. 

Link to Trello card
https://trello.com/c/C4Pd8KFG/225-bug-errors-generated-when-viewing-an-empty-list-in-secure-message-the-parameter-id-is-required-s

**How to review**
Run the tests/acceptance tests and when checking a thread list with no messages, no exception should be generated on secure-message and party-service

### DIFF
--- a/secure_message/common/utilities.py
+++ b/secure_message/common/utilities.py
@@ -148,6 +148,9 @@ def add_business_details(messages):
 
 def add_users_and_business_details(messages):
     """Add both user and business details to messages based on data from party service"""
+    if not messages:
+        return messages
+
     messages = add_to_details(messages)
     messages = add_from_details(messages)
     logger.info("Successfully added to and from details")


secure_message/common/utilities.py
```diff
@@ -148,6 +148,9 @@ def add_business_details(messages):
 
 def add_users_and_business_details(messages):
     """Add both user and business details to messages based on data from party service"""
+    if not messages:
```
shahi645: Really small change, but its better standard to have `if messages` then the `else` statement is the `if not`:
```
if messages:
	messages = add_to_details(messages)
    messages = add_from_details(messages)
    logger.info("Successfully added to and from details")
    messages = add_business_details(messages)
    logger.info("Successfully added business details")
    return messages
return messages
```

shahi645: Also is this tested anywhere?

ryangrundy7: I can add a test no problem 👍 I always thought it was better to do ```if not``` but I'm up for changing it

ajmaddaford: I don't think we should do it like this. We should enforce that `add_users_and_business_details` is passed messages. The only place this method is called when there may be no messages is in the `ThreadList` resource in threads.py. We should check there if messages is `None` before attempting to call `add_users_and_business_details`. 

This method should have a check for when `messages` is `None` and raise an exception, something like:

```python
def add_users_and_business_details(messages):
	if messages is None:
		raise ValueError('messages is required')
	
	... existing method ...
```

ajmaddaford: I don't think we should do it like this. We should raise an exception here if no messages are passed, something like:
```python
def add_users_and_business_details(messages):
	if messages is None:
		raise ValueError('messages is required')

	... existing method ...
```

The only place `add_users_and_business_details` is called when messages may be `None` is from the `ThreadList` resource in threads.py. We should deal with it there and check we have messages before calling `add_users_and_business_details`. The other places where the method is called is in the context of a conversation, where there will always be at least one message.

tests/app/test_utilities.py
```diff
@@ -348,3 +349,9 @@ def test_add_to_details_with_internal_message_first(self):
 
             self.assertEqual(result[0], populated_message[0])
             self.assertEqual(result[1], populated_message[1])
+
+    def test_add_users_and_business_details_empty_messages_raises_Value_Error(self):
+        messages = None
+
+        with self.app.app_context(), self.assertRaises(ValueError):
```
ajmaddaford: I don't think you need an app_context to test this?

ryangrundy7: Will remove it 👍

secure_message/common/utilities.py
```diff
@@ -148,6 +148,8 @@ def add_business_details(messages):
 
 def add_users_and_business_details(messages):
     """Add both user and business details to messages based on data from party service"""
+    if messages is None:
+        raise ValueError('messages is required')
```
gavinleeedwards: It shouldn't ever be hit if messages=None but if it is then it generates a valueError at exception level with a large stack trace and no context as to what caused it (party_uuid, survey_ref etc), which is what we're trying to move away from.  At this point, it would be a genuine error but could you add some context around it to help diagnose any issues and maybe not dump out the trace.

`{“error”: “ValueError(‘messages is required’,)“, “level”: “exception”, “service”: “ras-secure-message”, “created_at”: “2018-07-19T13:381532003935"}
Traceback (most recent call last):.....`

shahi645: Good catch, @ryangrundy7 worth putting some error logging here, with some key to identify, maybe `url` and `status code` to

ryangrundy7: Good shout Gav 👍

insacuri: The error message could be improved.  If you're looking through the logs and saw 'messages is required'  you'd think that the persons grammar skills are weak.  Once you look at the code you can see why it was written that way, but you want to make log lines reasonably expressive where possible.  Something like 'Cannot add details to a NoneType object'.
As I'm writing this,  I think the line above should be `if not messages:` because what happens if you give this an empty list?  In that case, then the error message could be something like 'There are no messages to process' (don't use that though because I don't quite think that's right).

insacuri: If you give it an empty list, `add_business_details` will make an empty call to the party service (not sure if that will error or just generate useless network traffic... both things are undesirable).  Other ones I think will do the same.  I definitely think you should do `if not messages:` to cover against both None and empty lists being passed to this function

shahi645: I think there is a heck before hand where this is called from, but i guess better safe than sorry

secure_message/resources/threads.py
```diff
@@ -109,7 +109,8 @@ def get():
 
         logger.info("Successfully retrieved threads for user", user_uuid=g.user.user_uuid)
         messages, links = process_paginated_list(result, request.host_url, g.user, message_args, THREAD_LIST_ENDPOINT)
-        messages = add_users_and_business_details(messages)
+        if messages:
```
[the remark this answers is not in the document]
ryangrundy7: @insacuri  Could be a good idea but I could also see it potentially polluting the splunk logs if the user keeps looking at empty lists.

secure_message/common/utilities.py
shahi645: also if its an empty list, `if messages` will evaluate to false

secure_message/resources/threads.py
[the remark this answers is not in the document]
insacuri: Don't worry about polluting the splunk logs, better to have too much logging and strip it back then not have enough when a problem occurs.  A middle ground option could be to log out a line saying 'No messages retrieved' or something to that effect and not bother logging out the args. I think we can get some useful url param information from another log line that gets logged out every time the service gets hit.  

Feel free to not bother with it though because it's not like we're logging 'Retrieved X messages/threads' for the non-zero cases, so that asymmetry might be weird.

secure_message/common/utilities.py
ryangrundy7: I think it's a good idea, better to be safe with it

tests/app/test_utilities.py
```diff
@@ -348,3 +349,8 @@ def test_add_to_details_with_internal_message_first(self):
 
             self.assertEqual(result[0], populated_message[0])
             self.assertEqual(result[1], populated_message[1])
+
+    def test_add_users_and_business_details_empty_messages_raises_Value_Error(self):
```
ajmaddaford: Method names should be all lower case - `_Value_Error` should be `_value_error`

tests/app/test_utilities.py
```diff
@@ -348,3 +349,8 @@ def test_add_to_details_with_internal_message_first(self):
...
+
+        with self.assertRaises(ValueError):
+            add_users_and_business_details(messages=None)
```
ajmaddaford: We should have a test for an empty list i.e. `[]` as well as `None`

PR #226, opened 2018-08-07
Remove calls to case service

gavinleeedwards:

https://trello.com/c/tVNXWOVU/267-remove-case-service-from-sm-s

Currently if a collection case id is present when a message is sent, specifically from frontstage, the Secure Messaging service will attempt to post a case event to the case service that a message has been sent. After the BI cases have been removed it has been decided tracking this event against a business case. 

**Changes**

- Removed all interaction with the case service from Secure Messaging
- Updated Cryptography to 2.3
- Fixed pylint issues

**How to review**

- Secure messaging should still work - do a `pip install --dev` to pick up the upgraded cryptography
- All tests (including acceptance tests related to SM) should still pass
- Check that there is no longer any interaction with the case service


secure_message/resources/messages.py
```diff
@@ -75,7 +75,6 @@ def _alert_listeners(message):
         """used to alert user and case service once messages have been saved"""
```
MIslam92: Remove case service from ` """used to alert user and case service once messages have been saved"""`

MIslam92:
Also there is still some stuff in the API.md around case service? So might be worth quickly remove it from there too

API.md
```diff
@@ -50,7 +50,7 @@ See the endpoint descriptions for detailed usage of each field. This is an overv
 * Subject . (subject) The subject of the message. Limited in the API to 100 characters , but since replies are prefixed with 'Re: ' then in practice it is 96 characters.
 * Body . (body) Up to 10000 characters.
 * Survey . (survey). This is the uuid of the survey . It is mandatory when saving a message.
-* Collection Case . (collection_case) uuid of the collection case. Can be used as a filter option (cc). Used to send to the case service to inform it that a new message has been sent on the case.
+* Collection Case . (collection_case) uuid of the collection case. Can be used as a filter option (cc).
```
gemmai95: I think you can remove this line completely.

gavinleeedwards: Removed the line and added a card to look at removing collection_case altogether from secure message after it's no longer being passed in.

gavinleeedwards:
@MIslam92 I knew I'd miss something!! Removed mentions of the case service in the docstring and API so can you re-review please? Ta.

PR #232, opened 2018-08-31
Put Zipkin trace IDs into log messages.

rjweeks:
Put Zipkin trace IDs into log messages


rjweeks:
[Add the Zipkin correlation and span ID into the logs for Python services (2d)](https://trello.com/c/5cYDTILb/156-add-the-zipkin-correlation-and-span-id-into-the-logs-for-python-services-2d)

Issue #133, opened 2018-01-29
Lack of context in log messages


necrophonic:
There seem to be a lot of log messages that are missing any kind of context, e.g:

https://github.com/ONSdigital/ras-secure-message/blob/2dc5b294875192fe79a1bcf82fdb2ed30cf2f423/secure_message/resources/messages.py#L54

There's no reference that can be tied back to any particular message and no way of tying this message to a subsequent error further down in the same method.

Logging without context (transaction id / collaboration id etc) is not useful for debugging and just emits noise into the logs.

insacuri:
Closing as we've improved this over the years.  It might not be perfect but it's better than it was in a lot of places.

PR #236, opened 2018-10-23
Improve logging for alerts

gemmai95:
Improve logging for alerting purposes.

Link to Trello card
Trello: https://trello.com/c/bdY6GFqq

**How to review**

### DIFF
--- a/secure_message/common/alerts.py
+++ b/secure_message/common/alerts.py
@@ -14,7 +14,7 @@ class AlertViaGovNotify:
     """Notify Api handler"""
 
     @staticmethod
-    def send(email, reference):
+    def send(email, reference, survey_id, party_id):
 
         notification = {
             "emailAddress": email,
@@ -27,18 +27,18 @@ class AlertViaGovNotify:
                                  json=notification)
 
         if response.status_code != 201:
-            raise RasNotifyException(code=500)
+            raise RasNotifyException(code=500, survey_id=survey_id, party_id=party_id)
         else:
             logger.info('Sent secure message email notification, via RM Notify-Gateway to GOV.UK Notify.',
-                        message_id=response.json()["id"])
+                        message_id=response.json()["id"], survey_id=survey_id, party_id=party_id)
 
 
 class AlertViaLogging:
     """Alert goes via gov notify (0) or via logs (1)"""
 
     @staticmethod
-    def send(email, reference):
-        logger.info(f'email details {email}, {reference}')
+    def send(email, reference, survey_id, party_id):
+        logger.info(f'email details {email}, {reference}, survey={survey_id} and party_id={party_id}')
 
 
 class AlertUser:
@@ -48,5 +48,5 @@ class AlertUser:
         if alerter is not None:
             self.alert_method = alerter
 
-    def send(self, email, reference):
-        self.alert_method.send(email, reference)
+    def send(self, email, reference, survey_id, party_id):
+        self.alert_method.send(email, reference, survey_id=survey_id, party_id=party_id)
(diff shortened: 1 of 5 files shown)


necrophonic:
Does this fix #133 ?

gemmai95:
@necrophonic I think it's a start to looking at that but I think it needs looked into more.

secure_message/common/alerts.py
```diff
@@ -27,18 +27,18 @@ def send(email, reference):
...
 
 
 class AlertViaLogging:
     """Alert goes via gov notify (0) or via logs (1)"""
 
     @staticmethod
-    def send(email, reference):
-        logger.info(f'email details {email}, {reference}')
+    def send(email, reference, survey_id, party_id):
+        logger.info(f'email details {email}, {reference}, survey={survey_id} and party_id={party_id}')
```
ryangrundy7: Might be a good idea to give keys to `email` and `reference` so its similar to the other key/values

secure_message/common/alerts.py
```diff
@@ -27,18 +27,18 @@ def send(email, reference):
...
 
 
 class AlertViaLogging:
     """Alert goes via gov notify (0) or via logs (1)"""
 
     @staticmethod
-    def send(email, reference):
-        logger.info(f'email details {email}, {reference}')
+    def send(email, msg_id, survey_id, party_id):
+        logger.info(f'email details email={email}, msg_id={msg_id}, survey={survey_id} and party_id={party_id}')
```
andrewmil: Do we want to log out email addresses? I'm sure in the past IA have asked us not to do that

gemmai95: This one is only used for tests but good point shouldn't be doing it anywhere

andrewmil: Also this should use key/value pair logging
`logger.info("Email sent", msg_id={msg_id}, survey={survey_id}, party_id={party_id})`

andrewmil: Ah yeah this is just when notify is disabled right? Might be fine then

secure_message/common/alerts.py
```diff
@@ -27,18 +27,18 @@ def send(email, reference):
                                  json=notification)
 
         if response.status_code != 201:
```
owencjones: Gateway returns `201` for a success?  As in temporary redirect?

NickGrantONS: 201 means created

secure_message/common/alerts.py
```diff
@@ -27,18 +27,18 @@ def send(email, reference):
...
-            raise RasNotifyException(code=500)
+            raise RasNotifyException(code=500, survey_id=survey_id, party_id=party_id)
```
AndrewTorrance: But the notify is only called on message post. And message post swallows all exceptions in _alert_listeners( The reason being that the message post was deemed a success even if the notify failed - business choice). So raising a 500 here is misleading as it will never get back to the caller, however it WILL lead to log entries . So its possibly worth a note to say that the exception is swallowed and the caller does not get a 500 in this circumstance

secure_message/common/alerts.py
```diff
@@ -48,5 +48,5 @@ def __init__(self, alerter=None):
         if alerter is not None:
             self.alert_method = alerter
```
AndrewTorrance: Comment only- No change needed: With only 2 send methods duplicating the logging is ok ish. However since the AlertViaLogging and AlertViaGovNotify are only plug ins for the AlertUser class we could log in AlertUser.send and then only log the send method in the plug in classes .  This would men that we would not need to pass params to the plug in classes simply for the logs .

PR #240, opened 2018-12-20
Message url now sent to notify gateway

ryangrundy7:
We want to change the notification email that the respondent gets when an internal message is received. To do this we need to send the conversation url that the respondent can use to access it. Once this is sent to notify gateway, it can be updated with the appropriate message.

[Trello](https://trello.com/c/nLsI8fjn)

**Describe what you have changed**
- Added a personalisation key/value to the notification send to notify gateway

**How to review**
- Quite hard to review, just make sure that the url being sent to notify gateway is correct. This could be done by debugging and seeing that the `MESSAGE_URL` in personalisation will lead you to the respondents message on frontstage.
[Frontstage pr](https://github.com/ONSdigital/ras-frontstage/pull/424)


config.py
```diff
@@ -61,6 +61,10 @@ class Config:
     RAS_PARTY_SERVICE_PROTOCOL = os.getenv('RAS_PARTY_SERVICE_PROTOCOL', 'http')
     RAS_PARTY_SERVICE = f'{RAS_PARTY_SERVICE_PROTOCOL}://{RAS_PARTY_SERVICE_HOST}:{RAS_PARTY_SERVICE_PORT}/'
 
+    FRONTSTAGE_SERVICE_HOST = os.getenv('RAS_FRONTSTAGE_HOST', 'localhost')
+    FRONTSTAGE_SERVICE_PORT = os.getenv('RAS_FRONTSTAGE_PORT', 8082)
+    RAS_FRONTSTAGE_SERVICE = f'http://{FRONTSTAGE_SERVICE_HOST}:{FRONTSTAGE_SERVICE_PORT}/'
```
gavinleeedwards: Will need to split out the protocol (like the Party URL).  Localhost uses http, where preprod will hang unless you use https.

secure_message/resources/messages.py
```diff
@@ -89,12 +89,19 @@ def _try_send_alert_email(message):
...
+                    alert_user.send(recipient_email, message.msg_id, personalisation=personalisation,
+                                    survey_id=message.survey, party_id=party_data[0]['id'])
                 else:
                     logger.error('User does not have an emailAddress specified', msg_to=message.msg_to[0])
             # else not testable as fails validation
         return party_data
 
+    @staticmethod
+    def create_message_url(thread_id):
+        url = f'{current_app.config["RAS_FRONTSTAGE_SERVICE"]}secure-message/thread/{thread_id}'
```
gavinleeedwards: Can you add the anchor to #latest-message

config.py
ryangrundy7: Added protocol but also changed the env names so they match with the variable names 👍

secure_message/resources/messages.py
ryangrundy7: Added anchor to the url

secure_message/resources/messages.py
```diff
@@ -89,12 +89,19 @@ def _try_send_alert_email(message):
...
-                    alert_user.send(recipient_email, message.msg_id, survey_id=message.survey, party_id=party_data[0]['id'])
+                    personalisation = MessageSend.create_message_url(message.thread_id)
+                    alert_user.send(recipient_email, message.msg_id, personalisation=personalisation,
+                                    survey_id=message.survey, party_id=party_data[0]['id'])
                 else:
                     logger.error('User does not have an emailAddress specified', msg_to=message.msg_to[0])
             # else not testable as fails validation
         return party_data
 
+    @staticmethod
```
AndrewTorrance: If its only used within MessageSend ( and there is no test ?) then perhaps its name should start be _create_message_url ?

ryangrundy7: Changed to `_create_message_url`

tests/behavioural/features/v2/message_post.feature
```diff
@@ -22,3 +22,11 @@ Feature: Message Send V2 Endpoint
     | user        |
     | specific user |
     | group        |
```
AndrewTorrance: This is in the wrong feature file , its testing the threads_get endpoint and should be in threads_get.feature

tests/app/test_message.py
```diff
@@ -273,6 +274,15 @@ def test_msg_to_validation_invalid_respondent(self):
 
         self.assertTrue(errors == {'msg_to': ['NotAValidUser is not a valid respondent.']})
 
+    def test_message_url_is_created_correctly(self):
+        self.threadId = 'threadid'
+        with self.app.app_context():
+            message_send = MessageSend()
+            response = message_send._create_message_url(self.threadId)
+
+        self.assertTrue(response == {'MESSAGE_URL': 'http://localhost:8082/secure-message/thread/'
```
AndrewTorrance: Minor point - may want to make the string the same f string as in the code . Then if values change the test still passes, but if the format of the url changes then the test fails.

tests/behavioural/features/v2/message_post.feature
ryangrundy7: My bad 👍

PR #241, opened 2019-01-02
Added my-conversations option to get threads

AndrewTorrance:

The requirement is to be able to have a my_conversations tab on response_ops ui . To support this the secure_message service get threads end point needs to support an option whereby an internal user can see 'their' conversations. Their conversations being ones where  message was sent directly 
to a specific user , or that user sent the last message.

Changes:- added my_conversations as optional parameter on get_threads endpoint.
Created an alternate internal user in the mock uaa
Changed the ids of the internal and alternate internal user to make db records easier to read 
Added several new scenarios for testing . No 'unit tests' added used behave tests to test outside in .

**How to review**
At this point with no ui , the only tests are the behave tests that validate the service. So in short- run the tests .


config.py
```diff
@@ -21,7 +21,7 @@ class Config:
     VERSION = os.getenv('VERSION', '0.11.0')
 
     SECURE_MESSAGING_DATABASE_URL = os.getenv(
-        'SECURE_MESSAGING_DATABASE_URL', 'postgresql://postgres:postgres@localhost:5432')
+        'SECURE_MESSAGING_DATABASE_URL', 'postgresql://postgres:postgres@localhost:6432')
```
ryangrundy7: Did you mean to change this?

AndrewTorrance:
More functionality identified so closing this PR to add in the new functionality around the count of my messages

PR #243, opened 2019-01-04
603 my conversations

AndrewTorrance:
The requirement is that response ops ui has the ability to have an additional tab whereby an internal user can select their messages. That being conversations where they were one of the actors in the latest message on that conversation. To support this two changes were needed in the secure message api:
1) The ability to pass a parameter ('my_conversations = true') in the get threads end point
2) The ability to pass a parameter ('my_conversations = true) in the get threads count end point

The first so that we can select the correct records , and the second so that the count ( used for screen display) tracks the correct record count. 
In doing this a number of tidy ups where done . Principally around incorrectly named steps . Piplock had to be regenerated due to security warnings.
Only internal users can access my_conversations and get a count of threads with my_conversations. Accessing them as a respondent would need additional changes, not required , and will result in an error.

Link to Trello card : https://trello.com/c/djNU7lM5

**How to review**

Either use postman to enter data or run the behave tests . Several have been added that cover the known scenarios. The additional parameters on both endpoints default to off , so old clients should not be affected.


API.md
```diff
@@ -281,12 +281,18 @@ This gives a count of messages that have not been read for a specific user. Ther
...
+This returns the latest message in each conversation(aka thread) that satisfies the passed in
+criteria.
 It can use the same filter arguments as Get Messages, and returns the latest message in each thread that satisfies
-the criteria passed in .
-[Get Message List](#get-message-list)
+the criteria passed in [Get Message List](#get-message-list)
+
+Additional parameters above those on Get Message List:  
+
+is_closed=true  if set then only returns conversations that have been marked as closed, else returns open
```
mfcrocker: I'd like to see these parameters in `code blocks`

API.md
```diff
@@ -444,6 +450,31 @@ Note V2 will have either uuids or 'GROUP' for the user ids, and a uuid for the s
 For descriptions of @msg_from, @msg_to and @ru see messages get
 [Get Message List](#get-message-list)
 
+## Get Conversations count  
+
+`GET /messages/count`
+
+This returns a count of the conversations that satisfy the search criteria. Available search criteria matches those on get threads:
+
+* survey . If set then only conversations regarding this survey will be considered. Defaults to all surveys
```
mfcrocker: As above, code blocks

secure_message/repository/retriever.py
```diff
@@ -117,6 +135,7 @@ def _retrieve_internal_thread_list(request_args, user):
             conditions.append(SecureMessage.collection_exercise == request_args.ce)
 
         try:
+
```
mfcrocker: 🎨

secure_message/repository/retriever.py
```diff
@@ -31,22 +31,40 @@ def unread_message_count(user):
...
-            survey_conditions.append(True)
+        if cc:
+            conditions.append(SecureMessage.collection_case == cc)
+
+        if ce:
+            conditions.append(SecureMessage.collection_exercise == ce)
 
         try:
-            result = SecureMessage.query.join(Conversation) \
+
```
mfcrocker: :art:

API.md
insacuri: Agreed.  And same for the other newly added params

secure_message/repository/retriever.py
```diff
@@ -31,22 +31,40 @@ def unread_message_count(user):
         return result
 
     @staticmethod
-    def thread_count_by_survey(survey, is_closed):
+    def thread_count_by_survey(surveys, is_closed, user, my_conversations=False, ru_id=None, cc=None, ce=None):
```
insacuri: Take a look at https://github.com/ONSdigital/ras-secure-message/blob/master/secure_message/resources/threads.py#L108 .  I think this function should be changed to operate in a similar way where you pass in an object (in ThreadList it's whatever get_options returns) that contains all the bits as a single variable instead of passing them in 1 by 1.
Pylint was complaining that there were 7 variables being passed in, and while I could've ignored it, I agree that this function has too many variables being passed in.

secure_message/resources/threads.py
```diff
@@ -113,16 +116,26 @@ def get():
             messages = add_users_and_business_details(messages)
         return jsonify({"messages": messages, "_links": links})
 
+    @staticmethod
+    def _validate_request(request_args, user):
+        if request_args.my_conversations and user.is_respondent:
+            logger.error('My conversations option not available to respondents')
```
insacuri:
```suggestion
            logger.error('My conversations option not available to respondents', user_uuid=user.user_uuid)
```

Should log out the user so if we need to follow up (say, a user is trying to probe our system, causing lots of errors), then we know who's doing it

secure_message/resources/threads.py
```diff
@@ -113,16 +116,26 @@ def get():
...
+        cc = request.args.get('cc', None)
+        ce = request.args.get('ce', None)
+        ru = request.args.get('ru_id', None)
 
         if not g.user.is_internal:
             logger.info("Thread count should be internal users only", user_uuid=g.user.user_uuid)
             abort(403)
 
-        return jsonify(total=Retriever.thread_count_by_survey(surveys, is_closed))
+        return jsonify(total=Retriever.thread_count_by_survey(surveys, is_closed, g.user, my_conversations, ru, cc, ce))
```
insacuri: See above with my comment about not passing everything through individually

tests/behavioural/features/steps/secure_messaging_context_helper.py
```diff
@@ -29,10 +29,10 @@ class SecureMessagingContextHelper:
     __DEFAULT_COLLECTION_EXERCISE = 'collection exercise1'
     __ALTERNATE_COLLECTION_EXERCISE = 'AnotherCollectionExercise'
 
-    __INTERNAL_USER_SPECIFIC_USER_ID = "SomeStringUntilWeGetIds"
+    __INTERNAL_USER_SPECIFIC_USER_ID = "SpecificInternalUserId"
     __INTERNAL_USER_GROUP = constants.NON_SPECIFIC_INTERNAL_USER
 
-    __ALTERNATIVE_INTERNAL_USER_SPECIFIC_USER_ID = "99951fcc-ed43-4cdb-ad1c-450f9986859b"
+    __ALTERNATIVE_INTERNAL_USER_SPECIFIC_USER_ID = "AlternateSpecificInternalUserId"
```
insacuri: I'm not a fan of this change.  I'd be interested in hearing peoples opinions but I look at tests sometimes to know what the inputs/outputs of a thing are, and what those inputs/outputs should roughly look like.  
I'd rather see a bunch of uuids for this and know that these ids should be uuids rather than some other random format.

AndrewTorrance: In the case of testing conversations having an indication of type in the db during tests makes tracking what is happening a lot easier. The data type used in the DB is var char not uuid . If the type was uuid then I would agree.

secure_message/resources/threads.py
```diff
@@ -113,16 +116,26 @@ def get():
...
 class ThreadCounter(Resource):
     """Get count of all open or closed messages"""
     @staticmethod
     def get():
         surveys = request.args.getlist('survey')
         is_closed = bool(strtobool(request.args.get('is_closed', 'False')))
+        my_conversations = bool(strtobool(request.args.get('my_conversations', 'False')))
+        cc = request.args.get('cc', None)
+        ce = request.args.get('ce', None)
+        ru = request.args.get('ru_id', None)
```
NickGrantONS: what do we do with this RU? I mean, do we have protection to stop people from guessing other RUs and seeing stuff they shouldn't?

secure_message/repository/retriever.py
AndrewTorrance: done

secure_message/repository/retriever.py
AndrewTorrance: done

secure_message/resources/threads.py
AndrewTorrance: Only an internal user hits this end point . On a respondent threads lookup the respondent uuid is a param in the search . They are free to guess an ru_id , its a uuid, but if they do then they only see their conversations.

tests/behavioural/features/steps/secure_messaging_context_helper.py
insacuri: Is it a varchar that always contains a uuid in preprod/prod?  or is it truly meant to contain whatever you want?

The question is mainly, was the db type NOT being uuid intentional?  If yes, then it's fine, but if not, then I don't think 'anything goes in the db' is a good enough reason.
As always, it's just an opinion, if you really want to leave it then I won't hold up this PR on something so minor ^^

secure_message/repository/retriever.py
AndrewTorrance: Added for consistency but feels a bit dirty, naming the specific params shows intent .

secure_message/resources/threads.py
AndrewTorrance: Added

tests/behavioural/features/steps/secure_messaging_context_helper.py
insacuri: Spoke about this, while I still think it should be changed, I'm not going to hold up the PR on something so minor.

PR #246, opened 2019-02-27
Change log level to info for some log lines

insacuri:

https://trello.com/c/QZQ1NQ73/780-improve-secure-messaging-logging-so-that-reasons-for-failure-to-retrieve-thread-can-be-determined-2

The logging in secure message isn't enough to diagnose all the issues that come from it.  This PR should increase both the amount of logging, and the quality of the logging to make future debugging easier.

**How to review**

### DIFF
--- a/secure_message/common/utilities.py
+++ b/secure_message/common/utilities.py
@@ -134,15 +134,21 @@ def add_from_details(messages):
     external_user_details = {}
     for user in party.get_users_details(get_external_user_uuid_list(messages)):
         external_user_details[user['id']] = user
+
+    msg_from = None
+    from_internal = None
+
     for message in messages:
         try:
-            if message["from_internal"]:
-                message.update({"@msg_from": internal_user_service.get_user_details(message["msg_from"])})
+            msg_from = message["msg_from"]
+            from_internal = message["from_internal"]
+            if from_internal:
+                message.update({"@msg_from": internal_user_service.get_user_details(msg_from)})
             else:
                 if external_user_details.get(message['msg_from']):
-                    message.update({'@msg_from': external_user_details.get(message['msg_from'])})
+                    message.update({'@msg_from': external_user_details.get(msg_from)})
         except IndexError:
-            logger.exception(f"Exception adding from details message:'{json.dumps(message)}'")
+            logger.exception("Exception adding from details message", msg_from=msg_from, from_internal=from_internal)
             raise
     return messages
 
--- a/secure_message/common/utilities.py
+++ b/secure_message/common/utilities.py
@@ -1,5 +1,4 @@
 import collections
-import json
 import logging
 import urllib.parse
 
--- a/secure_message/common/utilities.py
+++ b/secure_message/common/utilities.py
@@ -87,29 +87,38 @@ def add_to_details(messages):
     won't be a @msg_to value returned in the payload.  The API documentation notes that these elements
     aren't guaranteed to be provided so we're not breaking the contract by doing this.
 
-    Note:28th Feb 2019. Try except blocks added to try and identify cause of IndexError seen in splunk logs
-    No functionality added, just log and raise.
+    Note: Several of these lines of code could be combined into a more succinct view, spreading them out
+    is deliberate so that log stack traces are better able to identify the cause of log errors
     """
+
     external_user_details = {}
 
-    try:
-        external_users = get_external_user_uuid_list(messages)
-    except IndexError:
-        logger.exception(f"Exception getting external user uuids for messages:'{json.dumps(messages)}'")
-        raise
+    external_users = get_external_user_uuid_list(messages)
 
     for user in party.get_users_details(external_users):
         external_user_details[user['id']] = user
 
     for message in messages:
+
+        msg_to = None
+        from_internal = None
+
         try:
-            if not message["from_internal"]:
-                message.update({"@msg_to": [internal_user_service.get_user_details(message["msg_to"][0])]})
+            msg_to = message["msg_to"][0]
+            from_internal = message["from_internal"]
+
+            if not from_internal:
+                msg_to_details = internal_user_service.get_user_details(msg_to)
+                message.update({"@msg_to": [msg_to_details]})
             else:
-                if external_user_details.get(message['msg_to'][0]):
-                    message.update({'@msg_to': [external_user_details.get(message['msg_to'][0])]})
+                msg_to_details = external_user_details.get(msg_to)
+                if msg_to_details:
+                    message.update({'@msg_to': [msg_to_details]})
+                else:
+                    logger.info("No details found for message to", msg_to=msg_to)
+
         except IndexError:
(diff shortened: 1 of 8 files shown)


secure_message/common/utilities.py
```diff
@@ -78,30 +79,44 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
+    Note:28th Feb 2019. Try except blocks added to try and identify cause of IndexError seen in splunk logs
+    No functionality added, just log and raise.
     """
     external_user_details = {}
-    for user in party.get_users_details(get_external_user_uuid_list(messages)):
+
+    try:
+        external_users = get_external_user_uuid_list(messages)
+    except IndexError:
+        logger.exception(f"Exception getting external user uuids for messages:'{json.dumps(messages)}'")
```
insacuri: This is probably a bad idea.  What's inside of those messages?  We're not allowed to log any personally identifiable information and I think this might contain some of that bad stuff

secure_message/common/utilities.py
```diff
@@ -78,30 +79,44 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
-            if external_user_details.get(message['msg_to'][0]):
-                message.update({'@msg_to': [external_user_details.get(message['msg_to'][0])]})
+        try:
+            if not message["from_internal"]:
+                message.update({"@msg_to": [internal_user_service.get_user_details(message["msg_to"][0])]})
+            else:
+                if external_user_details.get(message['msg_to'][0]):
+                    message.update({'@msg_to': [external_user_details.get(message['msg_to'][0])]})
+        except IndexError:
+            logger.exception(f"Exception adding to details message:'{json.dumps(message)}'")
```
insacuri: Same as the other message.  Also, thinking about it, you don't want to dump out a giant blob of json in the logs as it's difficult to look at.  We should target the things we might care about and increase it (probably in a new field for easier searching) if it's not enough

secure_message/common/utilities.py
```diff
@@ -113,11 +128,15 @@ def add_from_details(messages):
...
-            if external_user_details.get(message['msg_from']):
-                message.update({'@msg_from': external_user_details.get(message['msg_from'])})
+        try:
+            if message["from_internal"]:
+                message.update({"@msg_from": internal_user_service.get_user_details(message["msg_from"])})
+            else:
+                if external_user_details.get(message['msg_from']):
+                    message.update({'@msg_from': external_user_details.get(message['msg_from'])})
+        except IndexError:
+            logger.exception(f"Exception adding from details message:'{json.dumps(message)}'")
```
insacuri: Here too

insacuri: Also also, we should be using fields for this information.  So this would become:
```python
logger.exception("Exception adding from details message", message=json.dumps(message))
```
But message is too broad, but you get the idea ^^

secure_message/repository/retriever.py
```diff
@@ -217,11 +217,11 @@ def _retrieve_thread_for_internal_user(thread_id):
...
 
             if not result.all():
-                logger.debug('Thread does not exist', thread_id=thread_id)
-                raise NotFound(description=f"Conversation with thread_id '{thread_id}' does not exist")
+                logger.error('Thread not retrieved for internal user', thread_id=thread_id)
+                raise NotFound(description=f"Conversation with thread_id '{thread_id}' not retrieved")
 
         except SQLAlchemyError:
-            logger.exception('Error retrieving conversation from database')
+            logger.exception(f"Error retrieving conversation from database thread_id '{thread_id}'")
```
insacuri: Use fields instead of fstrings.  Makes it easier to search for in splunk

secure_message/common/utilities.py
AndrewTorrance: This is a fishing expedition which is why it is so broad , but will rethink in light of not logging personal data

secure_message/common/utilities.py
AndrewTorrance: Good catch , the problem is I suspect that missing data may be causing this some of which is personally identifiable . This needs a rethink..

secure_message/common/utilities.py
```diff
@@ -78,30 +79,53 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
+    Note: Several of these lines of code could be combined into a more succinct view, spreading them out
+    is deliberate so that log stack traces are better able to identify the cause of log errors
     """
+
     external_user_details = {}
-    for user in party.get_users_details(get_external_user_uuid_list(messages)):
+
+    external_users = get_external_user_uuid_list(messages)
+
+    for user in party.get_users_details(external_users):
```
insacuri: Why split this over 2 lines if we're not going to add any extra logging?

secure_message/common/utilities.py
```diff
@@ -78,30 +79,53 @@ def process_paginated_list(paginated_list, host_url, user, message_args, endpoin
...
+
+            if not from_internal:
+                msg_to_details = internal_user_service.get_user_details(msg_to)
+                message.update({"@msg_to": [msg_to_details]})
+            else:
+                msg_to_details = external_user_details.get(msg_to)
+                if msg_to_details:
+                    message.update({'@msg_to': [msg_to_details]})
+                else:
+                    logger.info("No details found for message to", msg_to=msg_to)
```
insacuri: This error message could be improved as it doesn't make a lot of sense without the context of the code (and I've never been a huge fan of msg_to, it doesn't really describe WHAT it contains all that well).  Something like 'No details found for the message recipient' is more clear.

secure_message/common/utilities.py
AndrewTorrance: Only so that any stack trace specifies the exact culprit. No functional difference.

insacuri:
Can't approve my own PR, but it looks good to me!

secure_message/common/utilities.py
```diff
@@ -113,11 +135,15 @@ def add_from_details(messages):
...
-            if external_user_details.get(message['msg_from']):
-                message.update({'@msg_from': external_user_details.get(message['msg_from'])})
+        try:
+            if message["from_internal"]:
+                message.update({"@msg_from": internal_user_service.get_user_details(message["msg_from"])})
+            else:
+                if external_user_details.get(message['msg_from']):
+                    message.update({'@msg_from': external_user_details.get(message['msg_from'])})
+        except IndexError:
+            logger.exception(f"Exception adding from details message:'{json.dumps(message)}'")
```
rhiGriff: These should be key values pairs to be consistent with the software engineering community standards: https://github.com/ONSdigital/software-engineer-community/blob/master/standards/logging.md

secure_message/repository/retriever.py
rhiGriff: Same above

secure_message/common/utilities.py
AndrewTorrance: oo. Good catch I took them out of one func and missed the other changing

PR #249, opened 2019-05-14
Validates user claim on message post by calling out to party

AndrewTorrance:
There was no validation that a user had a claim on a survey/ru. Which meant that if a valid user had a url they could send a message in a conversation that they should not be able to. This change validates the claim prior to being able to post a message .

Link to Trello card

This change is the second attempt. The first used claims passed in the JWT, however it was discovered that this failed at a level of survey/ru combinations that was not significantly higher than that in production, and although it was a cleaner design it could be an insidious bug. Therefore it was decided to add an end point to party and have secure message validate the claim prior to message post. Other endpoints did not need such protection, get for example uses the user id as part of the query. 

Having decided to use party , the decision was then how to decide. Party already exposed a get which returned all the associations. We could have walked these . However that would have meant that knowledge about what constituted a valid or invalid claim would then be in secure message , so it would would be a separation of concerns issue and led to more coupling between the services. Since this is only being called on message post which is a relatively infrequent event it was decided to provide a new end point in party purely to validate a claim. 

**How to review**
Run the unit and behave tests . This has added extra functionality to the message post which is covered by tests in this repo and it only returns status codes that it already returned so there should be no need to create an acceptance test for this. However it is possibly worthwhile confirming this by running the acceptance tests

### DIFF
--- a/secure_message/resources/messages.py
+++ b/secure_message/resources/messages.py
@@ -45,15 +45,26 @@ class MessageSend(Resource):
         post_data['from_internal'] = g.user.is_internal
         message = self._validate_post_data(post_data)
 
-        if message.errors == {}:
-            logger.info("Message passed validation")
-            self._message_save(message)
-            # listener errors are logged but still a 201 reported
-            MessageSend._alert_listeners(message.data)
-            return make_response(jsonify({'status': '201', 'msg_id': message.data.msg_id, 'thread_id': message.data.thread_id}), 201)
+        if message.errors != {}:
+            logger.error('Message send failed', errors=message.errors)
+            return make_response(jsonify(message.errors), 400)
 
-        logger.error('Message send failed', errors=message.errors)
-        return make_response(jsonify(message.errors), 400)
+        # Validate claim
+        if not self._has_valid_claim(g.user, message.data):
+            logger.error("Message send failed", error="Invalid claim")
+            return make_response(jsonify("Invalid claim"), 400)
+
+        logger.info("Message passed validation")
+        self._message_save(message)
+        # listener errors are logged but still a 201 reported
+        MessageSend._alert_listeners(message.data)
+        return make_response(jsonify({'status': '201', 'msg_id': message.data.msg_id, 'thread_id': message.data.thread_id}), 201)
+
+    @staticmethod
+    def _has_valid_claim(user, message):
+        """Validates that the user has a valid claim to interact with the business and survey in the post data
+        internal users have claims to everything, respondents need to check against party"""
+        return user.is_internal or party.does_user_have_claim(user.user_uuid, message.ru_id, message.survey)
 
     @staticmethod
     def _message_save(message):
--- a/secure_message/resources/messages.py
+++ b/secure_message/resources/messages.py
@@ -52,7 +52,7 @@ class MessageSend(Resource):
         # Validate claim
         if not self._has_valid_claim(g.user, message.data):
             logger.error("Message send failed", error="Invalid claim")
-            return make_response(jsonify("Invalid claim"), 400)
+            return make_response(jsonify("Invalid claim"), 403)
 
         logger.info("Message passed validation")
         self._message_save(message)
--- a/secure_message/resources/messages.py
+++ b/secure_message/resources/messages.py
@@ -45,7 +45,7 @@ class MessageSend(Resource):
         post_data['from_internal'] = g.user.is_internal
         message = self._validate_post_data(post_data)
 
-        if message.errors != {}:
+        if message.errors:
             logger.error('Message send failed', errors=message.errors)
             return make_response(jsonify(message.errors), 400)
 
--- a/secure_message/services/party_service.py
+++ b/secure_message/services/party_service.py
@@ -35,6 +35,36 @@ class PartyService:
         return self._get_user_details_from_party_service(uuids)
 
     @staticmethod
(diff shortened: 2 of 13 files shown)


secure_message/resources/messages.py
```diff
@@ -45,15 +45,26 @@ def post(self):
...
-            self._message_save(message)
-            # listener errors are logged but still a 201 reported
-            MessageSend._alert_listeners(message.data)
-            return make_response(jsonify({'status': '201', 'msg_id': message.data.msg_id, 'thread_id': message.data.thread_id}), 201)
-
-        logger.error('Message send failed', errors=message.errors)
-        return make_response(jsonify(message.errors), 400)
+        if message.errors != {}:
+            logger.error('Message send failed', errors=message.errors)
+            return make_response(jsonify(message.errors), 400)
```
mfcrocker: Is this definitely due to a bad request and not a server error? Would suggest 500 is more valid here (if not being more specific with our error handling than `{}`!!). There's also an argument that the Invalid claim should be 404 or 422 but it seems like there's been a move away from 400 indicating bad syntax of late :)

secure_message/services/party_service.py
```diff
@@ -34,6 +34,36 @@ def get_users_details(self, uuids):
...
+        try:
+            response.raise_for_status()
+        except HTTPError:
+            logger.exception('Failed to validate claim data with party service',
+                             status_code=response.status_code,
+                             text=response.text,
+                             respondent_id=user_id,
+                             bus_id=bus_id,
+                             survey_id=survey_id)
+            return []
```
mfcrocker: May be my lack of Python showing but it seems odd to return `[]` here rather than `False`

secure_message/resources/messages.py
```diff
@@ -45,15 +45,26 @@ def post(self):
...
+
+        # Validate claim
+        if not self._has_valid_claim(g.user, message.data):
+            logger.error("Message send failed", error="Invalid claim")
+            return make_response(jsonify("Invalid claim"), 400)
```
ryangrundy7: This should probably be a 403

secure_message/resources/messages.py
```diff
@@ -45,15 +45,26 @@ def post(self):
...
-        if message.errors == {}:
-            logger.info("Message passed validation")
-            self._message_save(message)
-            # listener errors are logged but still a 201 reported
-            MessageSend._alert_listeners(message.data)
-            return make_response(jsonify({'status': '201', 'msg_id': message.data.msg_id, 'thread_id': message.data.thread_id}), 201)
-
-        logger.error('Message send failed', errors=message.errors)
-        return make_response(jsonify(message.errors), 400)
+        if message.errors != {}:
```
insacuri: Can be simplified to `if message.errors`

tests/app/test_app.py
```diff
@@ -18,10 +17,13 @@
 from secure_message.api_mocks.party_service_mock import PartyServiceMock
 
 
-class FlaskTestCase(unittest.TestCase):
+class AppTestCase(unittest.TestCase):
     """Test case for application endpoints"""
 
     BRES_SURVEY = "33333333-22222-3333-4444-88dc018a1a4c"
+    SPECIFIC_INTERNAL_USER = "SpecificInternalUserId"
```
insacuri: Can this not be a real uuid as well if we're aliasing it anyway?

tests/app/test_app.py
```diff
@@ -67,24 +72,25 @@ def setUp(self):
...
         data = {'msg_to': ['0a7ad740-10d5-4ecb-b7ca-3c0384afb882'],
-                'msg_from': 'ce12b958-2a5f-44f4-a6da-861e59070a31',
+                'msg_from': AppTestCase.SPECIFIC_INTERNAL_USER,
                 'subject': 'MyMessage',
                 'body': 'hello',
                 'thread': "?",
                 'ru_id': 'f1a5e99c-8edf-489a-9c72-6cabe6c387fc',
                 'create_date': datetime.now(timezone.utc),
-                'read_date': datetime.now(timezone.utc)}
+                'survey': 'a-survey-id'}
```
insacuri: Is this a uuid or a number (stocks has a survey_id of 017 for instance)?  Can we have a real looking value here?

secure_message/resources/messages.py
insacuri: I think 400 seems about right.  Even if it is wrong, the code hasn't changed but moved as the logic is a bit different.  I'd leave it as is, as something might be relying on it being 400 specifically and we can raise a card to sort it if it turns out that it's wrong.

secure_message/services/party_service.py
insacuri: Agreed, the comment said True and False return values and the code below returns a true.  All sorts of messes happen when you have functions that return different types!

AndrewTorrance: Agreed , modifying

secure_message/resources/messages.py
AndrewTorrance: Changed

tests/app/test_app.py
AndrewTorrance: It lines up with the values in the mocks and leaving them human readable makes it easier while debugging

tests/app/test_app.py
AndrewTorrance: In use it is a uuid but is not tested here so it could be anything

PR #250, opened 2019-05-21
403 forbidden access to secure message

ChristopherMShelton:
Made change to frontstage to show 403 error when accessing message that does not belong to the user. Had to make the change to return a 403 rather than a 404 in secure message.

https://trello.com/c/iWD6OxY9/824-403-message-thread-not-found-status-page-3

**How to review**
 runtests in secure message


ChristopherMShelton:
> Just noticed the failing travis test. That needs to be fixed before this can go in.

fixed now was just a typo in the cucumber script

PR #254, opened 2019-06-17
Remove v2 endpoints

insacuri:
NOTE:  Do not merge yet.  This is quite a big PR so I thought I'd start the reviewing process early.
https://trello.com/c/2APRhkSh/911-remove-the-v2-definition-of-end-points-in-secure-message-5

A while ago we were going to version the API.  We've now decided against that as we control it.  This PR removes all the v2 endpoints so services using this api only have a single endpoint to hit (/messages, /threads, etc)

**How to review**

- Unit tests pass
- Acceptance tests pass
- Spins this up locally with ras-frontstage and response-operations-ui on the same named branch.  Messaging should still work as expected.


API.md
```diff
@@ -483,25 +480,24 @@ ce              If set then restricts the conversations to a specific collction
 ```
 ## Get Conversation by Id
 
-`GET /thread/{thread_id} or /v2/threads/<thread_id>`
+`GET /threads/<thread_id>`
```
jonnyshaw89: Why the change to plural?

jonnyshaw89: ah ok, this isn;t a change for this PR. it just being made consistent

PR #256, opened 2019-07-08
Getting rid of warnings

anwilkie:

The purpose of this PR is to get rid of the warnings that come up when the tests are run for secure message. One of the warnings that would come up multiple times relates to Marshmallow not being updated to v3.0, which means that a schema's `strict` variable is set to `none`. In Marshmallow v3.0, `strict` will always be `true` by default, so we needed to make sure that the tests still pass when `strict` is `true`.

Link to Trello card

[Trello card](https://trello.com/c/Q1v90aAS)

* Changed the parameters used for the `test_logging_message_endpoint` test in test_logging.py so that the test does not fail when `strict` is set to `true`.
* Changed `assertEquals` to `assertEqual`. (Seems this change was already part of the master branch, so it doesn't show up as being changed)

**How to review**

* In your IDE, find External Libraries -> Pipenv -> Python 3.6 -> site-packages -> marshmallow -> schema.py.
* On line 342, change `strict = None` to `strict = True`. The purpose of doing this is to simulate marshmallow 3.0. 
* Comment out or delete lines 719 and 720 (the `if self.strict:` statement). Since `strict` will always be `true` by default, you don't need to check for it later in the code, and leaving these lines in breaks some of the tests.
* Run the tests. All of them should pass.

As things are, this PR does not fully fix the problem (since the only way to resolve the marshmallow warnings is to edit the Python 3.6 environment files). Hopefully, some time in the future, this service will be updated to marshmallow 3.0, and the environment files can be updated accordingly.


insacuri:
@anwilkie can we not just update marshmallow to 3.X by just doing `pipenv lock` and making sure everything still works?

anwilkie:
Hold on. There's a problem with my commits. I need to fix it before anyone should review my changes.

anwilkie:
Okay, now it's ready to be reviewed.

tests/app/test_logging.py
```diff
@@ -24,10 +25,10 @@ def test_logging_message_endpoint(self):
         """logging message endpoint"""
         out = StringIO()
         sys.stdout = out
-        message = {'msg_to': 'richard', 'msg_from': 'torrance', 'subject': 'hello', 'body': 'hello world',
-                   'thread_id': ''}
+        message = {'msg_to': [constants.NON_SPECIFIC_INTERNAL_USER], 'msg_from': 'Gemma', 'subject': 'hello', 'body': 'hello world',
+                   'thread_id': '', 'ru_id': 'f1a5e99c-8edf-489a-9c72-6cabe6c387fc', 'survey': 'cb8accda-6118-4d3b-85a3-149e28960c54'}
```
necrophonic: Are these values arbitrarily chosen or are they based on other data? Might be worth defining them as constants (at least within the file) so we know if the values themselves are important to the test, or whether just a non-None value is acceptable.

anwilkie: Certain values would make the test fail when strict was set to true. I had to change the `msg_to` and `msg_from` values because it kept displaying an error message about how I didn't have the permission to enter `Richard` and `Torrance` as values. I had to try a whole bunch of different values (which I got from looking at what values were passed into other tests) until I found something that worked. `thread_id` and `survey` simply needed a valid non-none value.

necrophonic: They're probably specific values that get set up as part of test setup. May be worth making them constants then otherwise someone will change them and wonder why the tests start failing.

anwilkie: Which values do you want me to set as constants?

necrophonic: I'd say probably the names (Gemma etc), the `ru_id` and `survey` uuids

PR #263, opened 2019-09-16
Retrieve new respondent conversations

AndrewTorrance:
The requirement is that internal users can see new conversations from respondents on a new tab.
Once replied to they disappear from the new tab, so that new conversation requests do not go unnoticed

Link to Trello card
https://trello.com/c/G1PNU5Qa

Both the get threads and get threads count end points needed to support a new parameter for those conversations where a respondent started a new conversation

**How to review**
Run the new feature tests . These exercise the new functionality


API.md
```diff
@@ -443,7 +443,7 @@ Note, the response will have either uuids or 'GROUP' for the user ids, and a uui
 For descriptions of @msg_from, @msg_to and @ru see messages get
 [Get Message List](#get-message-list)
 
-## Get Conversations count  
+h2: Get Conversations count
```
insacuri: I know this is super picky, but we use ## for a second level heading in every other repo, could we keep it consistent?

AndrewTorrance: sure

API.md
```diff
@@ -443,7 +443,7 @@ Note, the response will have either uuids or 'GROUP' for the user ids, and a uui
 For descriptions of @msg_from, @msg_to and @ru see messages get
 [Get Message List](#get-message-list)
 
-## Get Conversations count  
+h2: ## Get Conversations count
```
insacuri: Also need to remove the h2 ^^

insacuri:
I think having 'my_conversations' and 'new_respondent_conversations' as boolean parameters is wrong.  I can't explain why (and I'm just doing some research on it now) but I think theres a cleaner way to do this.
One of my initial ideas could be to have it instead of being an explicit flag that you pass, you can instead have a length (though maybe that's wrong...) param so you can do`/threads?length=X`.
When combined with another param (`?started_by=internal/external` as a rough idea) you can get all the threads that have 1 message in them, and were started by an external user.  That would get you the same detail and be usable by other calls in the future.

API.md
AndrewTorrance: It isnt there ...... The md viewer is adding it

insacuri:
Code looks good, just need to test it.

PR #268, opened 2019-11-13
Cannot reply to deleted message

insacuri:

It's not possible to reply to deleted users.  A similar change has been made in response-operations-ui, but we need to do a check here incase the client gets around this restriction.  

https://trello.com/c/3TkwE4E8

**How to review**
Follow the 'test additional logic' part of the guide in https://github.com/ONSdigital/response-operations-ui/pull/371 Testing that will test this as secure-message is the part that generates the 404

### DIFF
--- a/secure_message/resources/messages.py
+++ b/secure_message/resources/messages.py
@@ -32,9 +32,6 @@ class MessageSend(Resource):
             logger.info('Request must set accept content type "application/json" in header.')
         post_data = request.get_json(force=True)
 
-        logger.info("Here!!!!!!!!!--------------!!!!!!!!!here")
-        logger.info(post_data)
-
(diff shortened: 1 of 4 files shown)


secure_message/resources/messages.py
```diff
@@ -35,6 +35,11 @@ def post(self):
         if 'msg_id' in post_data:
             raise BadRequest(description="Message can not include msg_id")
         if post_data.get('thread_id'):
```
AndrewTorrance: This is validating the message to field ,I think it should be part of  MessageSchema.validate_to instead of inline ?

insacuri: Agreed.  Also moved a bunch of other things in there too that should've always been there

PR #270, opened 2019-11-19
Rename ru_id to business_id

insacuri:
# Motivation and Context
We're changing ru_id to business_id to make it a bit more obvious what it is.  In doing this change we've had to change the contents of the payload from @ru_id to @business_details to keep the name up to date (and make the name more accurate as it doesn't only contain the id).
As part of the cleanup of the test names, it was discovered there was a lot of unused code so that's been deleted as well.

# What has changed
- All instances of @ru_id in the payload have been changed to @business_details
- A lot of unused behave test code was deleted
- A small script to update the database has been added
- The model has been updated with the new column name

# How to test?
- Test script works against your database.

Assuming your running this in docker:
- Purge your postgres container/image, rebuild it.  The model update should mean the column names are correct.
- Check the `idx_ru_survey_cc` index points to the correct columns (shouldn't mention ru_id anymore and should have business_id).

- run the unit and acceptance tests
- Try creating a few messages from both sides, it should all still work as before.

# Links
https://trello.com/c/nct0vF5E

# Screenshots (if appropriate):


tests/behavioural/features/steps/ru_field.py
```diff
@@ -1,31 +1,8 @@
-import nose.tools
-from behave import given, then, when
```
AndrewTorrance: Filename should change to business_id_field.py ?

PR #275, opened 2020-01-06
Return multiple counts

AndrewTorrance:

https://trello.com/c/3Bj1gYTy
There is a need to display the counts of secure message conversations in the title of each tab .
This PR enables that through a call to Secure message. 
It adds support for all_conversation_type parameter in the Thread count end point. This returns counts for open,closed,new_respondent_conversations and my conversations all in one endpoint. Its not practical to fulifil this in a single database query, a single conversation can count in multiple classes, and the selection criteria are complex. This code submits 4 queries and amalgamates the data .  

**How to review**
Run tests. New tests added for the additional parameter.

### DIFF
--- a/secure_message/repository/retriever.py
+++ b/secure_message/repository/retriever.py
@@ -85,7 +85,7 @@ class Retriever:
 
         totals = {}
 
-        args = set_conversation_type_args(request_args, is_closed=False)  # is_closed shown simply for clarity
+        args = set_conversation_type_args(request_args)  # is_closed defaults to False
         totals['open'] = Retriever.thread_count_by_survey(args, user)
 
         args = set_conversation_type_args(request_args, is_closed=True)
--- a/secure_message/repository/retriever.py
+++ b/secure_message/repository/retriever.py
@@ -10,6 +10,7 @@ from werkzeug.exceptions import InternalServerError, NotFound, Forbidden
 from secure_message.constants import NON_SPECIFIC_INTERNAL_USER
 from secure_message.common.eventsapi import EventsApi
 from secure_message.common.labels import Labels
+from secure_message.common.utilities import set_conversation_type_args
 from secure_message.repository.database import Conversation, db, Events, SecureMessage, Status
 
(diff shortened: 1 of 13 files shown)


secure_message/repository/retriever.py
```diff
@@ -74,6 +75,30 @@ def thread_count_by_survey(request_args, user):
...
+    def thread_count_by_survey_and_conversation_states(request_args, user):
+        """Return 4 conversation counts. They are for open, closed, my conversations and
+        new_respondent_conversation.
+        Given each of these counts uses different clauses to define them and they are not mutually exclusive,
+        they are difficult to achieve with the current db structure in a single query, hence this submits 4 db queries.
+        """
+
+        totals = {}
+
+        args = set_conversation_type_args(request_args, is_closed=False)  # is_closed shown simply for clarity
```
insacuri: Don't bother with the is_closed bit here.  It defaults to false in the function so this is superfluous

secure_message/resources/threads.py
```diff
@@ -127,7 +127,24 @@ def _validate_request(request_args, user):
 
 
 class ThreadCounter(Resource):
-    """Get count of all open or closed messages"""
+    """Get count of all conversations for a specific internal user
+    Typically for a specific survey, supports filtering by case, collection exercise, business party id etc
+    Args:
```
insacuri: I know this is picky but throughout all of our other repos we use the sphinx style of docstrings <https://www.datacamp.com/community/tutorials/docstrings-python#fifth-sub> could you convert it to that instead?

secure_message/repository/retriever.py
AndrewTorrance: Changed,

secure_message/resources/threads.py
AndrewTorrance: Disagree . These are not specific parameters so they are not suitable for the normal :param format. What I have done is to move the comments to the get_options function. But even then they are not individual params

PR #303, opened 2020-05-29
restored pipenv check to Makefile

jdfox:
Uncommented pipenv check that was toggled off in previous hotfix for pipenv bugs.

Also updated NumPy to fix critical vulnerability: https://nvd.nist.gov/vuln/detail/CVE-2019-6446

Link to Trello card
https://trello.com/c/3ZxIpfsX

**How to review**

pipenv check was previously commented out in https://github.com/ONSdigital/ras-secure-message/pull/297


jdfox:
Travis appears to be using its own NumPy version 1.13.3, and ignoring the Pipfile.lock update to 1.18.4. Might have to park this PR until we move ras-secure-message to Github Actions.

.travis.yml
```diff
@@ -10,6 +10,7 @@ cache:
 
 install:
     - pip install pipenv
+    - pip install --upgrade numpy
```
insacuri, line 13: Why do you have to upgrade it here?  Isn't it being upgraded in the pipfile.lock enough?

jdfox: Unfortunately not: Travis ignores the NumPy version in the lockfile. Luckily this is only a temporary workaround until ras-secure-message migrates to GitHub Actions.

insacuri: Ah that'll be fine then.  We're replacing travis with github actions soon enough anyway so this workaround is fine.

insacuri: Actually ignore that.  Can you try and upgrade the OS image it uses?  That numpy addition to pipfile will probably get forgotten about.

jdfox: John and Warren removed Travis config just now, so the problem has evaporated. :-)

Pipfile
```diff
@@ -22,6 +22,7 @@ retrying = "*"
 maya = "*"
 PyJWT = "*"
 requestsdefaulter = "*"
+numpy = "*"
```
insacuri, line 25: If we dont have numpy issues anymore, can we remove this?

jdfox: Unfortunately not yet: python-slim 3.6 and 3.8 support Debian Buster, but don't yet support Debian Bullseye, the one with a NumPy > 1.1.16.

PR #310, opened 2020-09-03
Notify pubsub

jterryons:
Remove the old notify stuff and add the new pubsub notify


insacuri (approved):
Looks good.  Remember to update everyones dev config!

PR #314, opened 2020-11-17
Mark all messages in a conversation as read

warrenbailey:
If an internal users sends two messages to an external respondent then it was impossible to mark the first as read so the user would always have a notification for it.

This change marks all messages in the thread as read when the latest one is.

https://trello.com/c/4PRWzCg3/234-ritm0459542-respondent-has-notification-of-unread-message-but-when-they-look-in-open-and-closed-there-are-no-new-messages

### DIFF
--- a/secure_message/repository/modifier.py
+++ b/secure_message/repository/modifier.py
@@ -82,8 +82,9 @@ class Modifier:
     def mark_message_as_read(message, user):
         """Remove unread label from status"""
         logger.debug("marking message as read with id", message=message['msg_id'])
+        # mark the message as read i.e. remove the unread label and set the `read at` time.
         Modifier._mark_read(message, user)
-        # also need to mark any other messages in the conversation as read
+        # also need to mark any other messages in the thread as read
         if message['thread_id']:
             try:
                 conversation = Retriever.retrieve_thread(message['thread_id'], user)
--- a/secure_message/repository/modifier.py
+++ b/secure_message/repository/modifier.py
@@ -102,6 +102,7 @@ class Modifier:
     def _mark_read(message, user):
         inbox = Labels.INBOX.value
         unread = Labels.UNREAD.value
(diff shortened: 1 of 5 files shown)


secure_message/repository/modifier.py
```diff
@@ -81,8 +81,29 @@ def add_unread(message, user):
     @staticmethod
     def mark_message_as_read(message, user):
         """Remove unread label from status"""
+        logger.debug("marking message as read with id", message=message['msg_id'])
+        # mark the message as read i.e. remove the unread label and set the `read at` time.
+        Modifier._mark_read(message, user)
+        # also need to mark any other messages in the thread as read
+        if message['thread_id']:
+            try:
+                conversation = Retriever.retrieve_thread(message['thread_id'], user)
```
mfcrocker, line 90: Do we have any way of only retrieving the messages that need acting on rather than pulling the whole thread? Understandable if we don't, but would be good to pull as little data as is necessary

mfcrocker (approved):
LGTM, one question

secure_message/repository/modifier.py
warrenbailey: not presently afaik but we could add it?

mfcrocker: Up to you, its not a prerequisite for merging this but it might be a better way of doing things

insacuri (approved):

secure_message/repository/modifier.py
warrenbailey: I think you're right, it should be a method on the Modifier. I'll take a look at how much work it would be to change. Ideally we'd tackle the whole service and make it more RESTful, it has a few oddities.

warrenbailey: I still have to pull the whole thread unfortunately, however I can change it to use a single db session to do all the work. I'll push up a change shortly.

insacuri (approved):
approving again for modified version

PR #335, opened 2021-04-16
add new column with default value and migration script

mybigopenbox:
# What and why?
secure messages to be routed to the correct teams, so that resourcing can be sufficiently arranged to cover all queries

# How to test?
deploy into environment and validate by viewing the database table

# Trello
S32 User account creation - Background functionality to route secure messages

### DIFF
--- a/_infra/helm/secure-message/Chart.yaml
+++ b/_infra/helm/secure-message/Chart.yaml
@@ -18,4 +18,4 @@ version: 1.1.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.1.16
+appVersion: 2.1.17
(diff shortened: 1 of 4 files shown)


warrenbailey (changes requested):
@mybigopenbox are the migration scripts missing from this PR as I can't see them?

Also for any new messages that are created we need to default the category to 'SURVEY' for now, I'd imagine that'll be done somewhere in the python code

mybigopenbox:
changes done.
two files 004_add_field_category_on_secure_message.sql and database.py within secure message both having a default value of 'SURVEY'

warrenbailey:
Great thanks 👍

warrenbailey (approved):

warrenbailey:
Approved, though it might be good to have a unit test. Then we can expand the tests once the new categories arrive.

warrenbailey (approved):

PR #336, opened 2021-04-21
Remove codecov

mybigopenbox:
# What and why?
Software package no longer required
# How to test?
Load in the environment
# Trello
S32 Remove codecov package use from python repos


mybigopenbox:
closed. produced another PR

PR #337, opened 2021-04-21
remove reference to codecov and update pipfile.lock file

mybigopenbox:
# What and why?
software packages not required
# How to test?
load into the environment
# Trello
S32 Remove codecov package use from python repos


mybigopenbox:
error relating to _row

PR #341, opened 2021-04-27
standardise fieldname on secure message

mybigopenbox:
# What and why?
standardise the naming convention in secure message service.

# How to test?
through the environments

# Trello
RASRM database rationalisation - Secure Message


warrenbailey:
This seems to be missing a chart version change?

warrenbailey:
This change looks good to me, but when I tested it in my dev environment I can't read a message in response operations and I can't send a new message in frontstage, on both occasions it generates an error.

warrenbailey (approved):

PR #363, opened 2021-09-29
Remove events table

insacuri:
# What and why?

The event table hasn't been needed for a while.  The info it stores is also stored against each message (as `sent_at` and `read_at`).  This PR removes all mention of it and updates the few places it was used to use the new columns.  This should simplify the design of this service greatly.

# How to test?

- Run acceptance tests in dev environment
- Create 2 more threads (internal to external, and the reverse)
- Deploy app to your environment
- Test it manually (check threads on both internal and external look as expected)
- After testing to make sure it works, delete the events table (with supplied script in PR).  Make sure it still works (manually and/or with acceptance tests)

Through my testing I saw that it worked just fine regardless of what state the database was in as long as the schema (and whatever tables) had been created.

# Trello

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,16 +7,16 @@ start:
 	pipenv run python run.py
 
 lint:
-	pipenv run flake8
 	pipenv check
 	pipenv run isort .
 	pipenv run black --line-length 120 .
+	pipenv run flake8
 
 lint-check:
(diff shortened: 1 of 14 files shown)


mfcrocker (approved):
LGTM, tested

PR #375, opened 2022-08-09
Return default user details when internal users not found

mfcrocker:
# What and why?
If an internal user has been deleted, any thread that comes from that user will generic 500 as the 404 from UAA isn't handled properly. This fixes it by just returning the default "ONS User" text that we already return if we end up with only partial information from UAA.

# How to test?

- Make a thread between an internal and external user
- Delete the internal user
- Check your message thread as an internal user; it shouldn't throw a 500 and should appear as being from "ONS User"

# Trello
https://trello.com/c/GOu2O5zI/1937-bug-handle-missing-internal-users-properly-in-secure-message-15m

### DIFF
--- a/secure_message/services/internal_user_service.py
+++ b/secure_message/services/internal_user_service.py
@@ -32,10 +32,12 @@ class InternalUserService:
             resp_json = response.json()
         except HTTPError:
             logger.exception("Failed to get user info", uuid=uuid)
-            raise
+            user_details = InternalUserService.get_default_user_details(uuid)
+            return user_details
         except ValueError:
             logger.exception("Failed to decode response JSON", uuid=uuid)
-            raise
+            user_details = InternalUserService.get_default_user_details(uuid)
+            return user_details
 
         try:
             user_details = {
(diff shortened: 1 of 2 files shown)


JacobHarrisonONS (approved):

insacuri (approved):

secure_message/services/internal_user_service.py
```diff
@@ -32,10 +32,12 @@ def get_user_details(uuid):
             resp_json = response.json()
         except HTTPError:
             logger.exception("Failed to get user info", uuid=uuid)
-            raise
+            user_details = InternalUserService.get_default_user_details(uuid)
+            return user_details
         except ValueError:
             logger.exception("Failed to decode response JSON", uuid=uuid)
```
LJBabbage, line 38: Might be worth extending the log lines to say its using default. ValueError is interesting

LJBabbage: I was trying to work it out and Jacob and Adam had approved

LJBabbage: This is fine for for a quick fix, but realistically this needs looking at again, if a major internal uses leaves and gets deleted our logs could be a mess with exceptions, we should handle it gracefully

PR #377, opened 2022-10-04
Fix vulnerable packages

insacuri:
# What and why?

There was a vulnerability in our packages.  This PR bumps everything possible to the latest version to fix it.  One thing that we couldn't bump past a certain version was Flask.  When it's set to =>2.2.x then the behave tests fail.  We can address this in another card specifically.

# How to test?

- Deploy app
- Run acceptance tests
- Use the secure message functionality yourself a bit.  It should all work

# Trello

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,7 +7,7 @@ url = "https://pypi.python.org/simple"
 python_version = "3.10"
 
 [packages]
-Flask = "*"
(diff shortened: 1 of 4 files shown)


JacobHarrisonONS (approved):
Tested in dev, secure message still works correctly. Send messages between and closed message.

PR #381, opened 2022-11-09
Upgrade to python 3.11 and action version upgrade

insacuri:
# What and why?

# How to test?

# Trello


matthew-robinson-ons (approved):
Upgrade functions correctly with acceptance tests passing.

PR #386, opened 2023-01-12
Flask upgrade

anwilkie:
# What and why?
Flask version used to be pinned; this PR unpins in and sorts out various test failures that result from upgrading.

Changes made to repo to maintain functionality and make tests pass:

Any instance of the tests calling the `database.db.init_app(current_app)` function was removed because the most recent version of Flask handles app initialisation by itself and doesn't require Werkzeug for that purpose. Hence, this line of code is unnecessary and was causing an error where it was trying to look for a `LocalProxy` Werkzeug object that didn't exist anymore.

The `paginate` method for flask-SQLAlchemy now requires keyword parameters rather than positional parameters, so any instance of this method now has the parameters explicitly defined.

The `get_engine` method for flask-SQLAlchemy doesn't accept an `app` parameter anymore, and the method itself is deprecated in the 3.0 versions of flask-SQLalchemy. Instead, you now simply call the `engine` object directly instead of using the getter method.

# How to test?

* Deploy to environment.
* Run acceptance tests.
* Go on frontstage/response-ops and send messages to make sure secure messaging works.

NOTE: When running unit tests, there is an SQLAlchemy deprecation warning relating to a pattern used in the tests involving SQL queries. This pattern is not compatible with version 2.0 of SQLAlchemy, but we have decided to ignore this for now because version 2.0 is still in beta. A card has been written up to deal with this warning in the future.

# Trello
[Card](https://trello.com/c/1M2cYmCa)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ lint:
 	pipenv run flake8
 
 lint-check:
-	#pipenv check -i 51668
+	pipenv check -i 51668
 	pipenv run isort --check-only .
 	pipenv run black --line-length 120 .
 	pipenv run flake8
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ lint:
 	pipenv run flake8
 
 lint-check:
-	pipenv check -i 51668
+	#pipenv check -i 51668
 	pipenv run isort --check-only .
 	pipenv run black --line-length 120 .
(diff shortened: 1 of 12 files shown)


matthew-robinson-ons (approved):
All looks good and secure message is behaving as expected.

PR #389, opened 2023-03-15
Upgrade sqlalchemy to 2.0

insacuri:
# What and why?

sqlalchemy 2.0 finally released so we upgraded to it to keep current.
It's not required too many changes as we used sessions and ORM models in a fairly simple way.  The bulk of the changes to get to 2.0 actually ended up being in the 'unit' tests (which are more integration tests) as they set up, tear down and check data in the database for each test.

# How to test?

- Deploy app to environment
- Run all 4 types of acceptance tests to ensure it works
- Try a few manual things (creating messages, closing messages, changing categories, etc)

To test it with a completely fresh database:

- delete the seccuremessage schema
- `kubectl rollout restart deployment secure-message`
- run the acceptance tests again to verify that it's happy
This will likely have a few errors in the logs as all the workers will try and run alembic, and a few might fail initially but given a few seconds it ends up sorting itself out.

# Trello


JacobHarrisonONS (approved):
Code looks good. Tested and no issues found.

PR #391, opened 2023-04-25
Update HPA to v2

LJBabbage:
# What and why?
Update autoscaling from v2beta2 to v2
# How to test?
The HPA change has already been proven here https://github.com/ONSdigital/ras-party/pull/389, but can be done again if required, just deploy via helm
# Trello


pricem14pc (approved):
Eyeballed only, but looks correct

PR #395, opened 2023-05-31
flask upgrade

anwilkie:
# What and why?
Flask upgrade

# How to test?
Deploy and run acceptance tests.

# Jira
[card](https://jira.ons.gov.uk/browse/RAS-784)

PR #398, opened 2023-06-22
Revert "flask upgrade"

anwilkie:
Reverts ONSdigital/ras-secure-message#395

This PR reverts the flask upgrade I made previously. I want to test if the technical message list being slow/erroring on preprod is the fault of the flask upgrade or something else.

I wanted to downgrade the flask version to 2.2.3 (what it was prior to upgrading), but Github actions fails the build because it reintroduces the security vulnerability. As such, I decided to pin the flask version to 2.2.5 (the oldest version that doesn't have the vulnerability). According to the [flask docs](https://flask.palletsprojects.com/en/2.3.x/changes), the only changes made between 2.2.3 and 2.2.5 are updates for werkzeug and the `Vary: Cookie` security fix

You can test this PR by deploying it and running acceptance tests.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,7 +7,7 @@ url = "https://pypi.python.org/simple"
 python_version = "3.11"
 
 [packages]
-Flask = "*"
+Flask = "==2.2.5"
 Flask-RESTful = "*"
 Flask-SQLAlchemy = "*"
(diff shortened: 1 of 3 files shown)


SteveScorfield (approved):
Looks good to me

PR #406, opened 2024-01-11
Making logging level configurable

pricem14pc:
# What and why?

Secure messaging should default to INFO and not DEBUG


LJBabbage (approved):
Although I haven't tested this fully with the config branch, I have pushed manually to my GCP and can confirm this works

PR #411, opened 2024-02-23
database error handling

arroyoAle:
# What and why?
Adds error handling around database errors and tests around this. Also updated the GHA workflow with the new `setup-gcloud` and replaces how we create the release
# How to test?
I haven't found a way to test this easily when deployed through spinnaker as other services fall over when there is a database connection issue (I cause this by deleting the postgres configmap). Instead I deploy the secure message PR locally and either insert `raise SQLAlchemyError` where it's being tested or then stop the local postgres deployment
- update uaa client secret so that secure message can connect to uaa on startup
- run local postgres, securemessage, rOps and/or frontstage
- make a SQLAlchemyError happen
- make sure that the whole stack trace is not printed to the logs
# Jira
[RAS-1031](https://jira.ons.gov.uk/browse/RAS-1031)

### DIFF
--- a/secure_message/repository/retriever.py
+++ b/secure_message/repository/retriever.py
@@ -14,6 +14,8 @@ from secure_message.repository.database import Conversation, SecureMessage, Stat
 
 logger = wrap_logger(logging.getLogger(__name__))
 
+THREAD_COUNT_ERROR = "Error retrieving count of threads by survey from database"
+
 
 class Retriever:
     """Created when retrieving messages"""
@@ -30,9 +32,12 @@ class Retriever:
                 .filter(Status.label == "UNREAD")
                 .count()
             )
+        except SQLAlchemyError as e:
+            logger.error(THREAD_COUNT_ERROR, error=e)
+            raise
         except Exception:
-            logger.exception("Error retrieving count of unread messages from database")
-            raise InternalServerError(description="Error retrieving count of unread messages from database")
+            logger.exception(THREAD_COUNT_ERROR)
+            raise InternalServerError(description=THREAD_COUNT_ERROR)
         return result
 
     @staticmethod
@@ -79,8 +84,11 @@ class Retriever:
 
             result = SecureMessage.query.filter(and_(*conditions)).distinct(SecureMessage.msg_id).count()
 
+        except SQLAlchemyError as e:
+            logger.error(THREAD_COUNT_ERROR, error=e)
+            raise
         except Exception as e:
-            logger.error("Error retrieving count of threads by survey from database", error=e)
+            logger.error(THREAD_COUNT_ERROR, error=e)
             raise InternalServerError(description="Error retrieving count of threads from database")
         return result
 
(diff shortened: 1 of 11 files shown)


secure_message/repository/retriever.py
```diff
@@ -14,6 +14,8 @@
 
 logger = wrap_logger(logging.getLogger(__name__))
 
+THREAD_COUNT_ERROR = "Error retrieving count of threads by survey from database"
```
mfcrocker, line 17: It's inconsistent to have pulled this message out but none of the other repeating ones throughout this PR. I'm personally not a fan of magic strings and would prefer the messages stay in the exceptions for readability purposes, but whichever way you choose could you choose the same one for all of them please :)

secure_message/repository/retriever.py
```diff
@@ -230,9 +238,12 @@ def retrieve_populated_message_object(message_id: str) -> SecureMessage:
             if result is None:
                 logger.info("Message ID not found", message_id=message_id)
                 raise NotFound(description=f"Message with msg_id '{message_id}' does not exist")
-        except SQLAlchemyError:
+        except SQLAlchemyError as e:
+            if e.__class__ == NoResultFound:
```
mfcrocker, line 242: Will this ever happen? I'd have thought this would be dealt with by `if result is None` above

secure_message/resources/messages.py
```diff
@@ -18,6 +20,7 @@
 
 logger = wrap_logger(logging.getLogger(__name__))
 
+MESSAGES_SERVICE_ERROR = "Messages service error"
```
mfcrocker, line 23: As above re: consistency. This is also not a super descriptive error message; if we can give any more detail we probably should.

mfcrocker (changes requested):
It works fine, just a couple of code style points to check out

secure_message/repository/retriever.py
arroyoAle: I've put them all back in with some additional information

secure_message/repository/retriever.py
arroyoAle: I thought so too but when I was testing it the database threw the NoResultFound instead of returning None when I provided it with a uuid that wasn't in the database

secure_message/resources/messages.py
arroyoAle: Updated as above

secure_message/resources/threads.py
```diff
@@ -18,6 +19,8 @@
 
 logger = wrap_logger(logging.getLogger(__name__))
 
+THREAD_SERVICE_ERROR = "Thread service error"
```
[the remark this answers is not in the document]
arroyoAle: Updated as above

mfcrocker (approved):

secure_message/repository/retriever.py
matthew-robinson-ons: Can this now be removed?

matthew-robinson-ons (commented):
Just the one query otherwise looks good

matthew-robinson-ons (approved):
Happy with these changes

PR #419, opened 2024-06-28
Have added bot signature, fixed CVE and ignored 2 CVEs

SteveScorfield:
# What and why?
Signed pr-bot commits and have ignored 2 vulnerabilities where there are currently no fixes for.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -24,7 +24,6 @@ numpy = "*"
 google-cloud-pubsub = "*"
 flask-sqlalchemy = "*"
 flask = "*"
-jinja2 = "*"
 
 [dev-packages]
 black = "*"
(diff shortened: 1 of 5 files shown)


Pipfile
```diff
@@ -24,6 +24,7 @@ numpy = "*"
 google-cloud-pubsub = "*"
 flask-sqlalchemy = "*"
 flask = "*"
+jinja2 = "*"
```
pricem14pc, line 27: We may do, but did you mean to add this as an explicit dependency? It's a third party dependency of `Flask-Cors`, `Flask-RESTful` and `flask-sqlalchemy`. I'm never sure if we should just upgrade those and let them resolve it.

SteveScorfield: This has been removed. I never meant to have included it. I was just testing a theory.

PR #421, opened 2024-07-24
remove second dockerfile

anwilkie:
# What and why?
Removing the dockerfile in the root directory to ensure consistency with other repos.

# How to test?
Check GA build

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-313)


Makefile
```diff
@@ -7,13 +7,13 @@ start:
 	pipenv run python run.py
 
 lint:
-	pipenv check -i 51668
+	pipenv check -i 70612 -i 71064 -i 70624
```
arroyoAle, line 10: again there is an update for requests (71064) can be a separate card

arroyoAle (approved):
Works fine just a little comment

Pipfile
```diff
@@ -24,6 +23,8 @@ numpy = "*"
 google-cloud-pubsub = "*"
 flask-sqlalchemy = "*"
 flask = "*"
+urllib3 = "*"
```
matthew-robinson-ons, line 26: we shouldn't be importing urllib3 it is managed as a subdependency of a number of packages, e.g. requests

matthew-robinson-ons (changes requested):
Just the one comment around urlib, the vulnerability is now fixed on main so after merging the import can be removed

PR #427, opened 2024-08-16
RAS-1265 Correct order of claim call in secure messaging

LJBabbage:
# What and why?
This will stop SM calling out to party unnecessarily 
# How to test?
Generate a message thread and make sure it still works
# Jira


arroyoAle (approved):
Looks good and messages still work in all three inboxes from rOps and frontstage

matthew-robinson-ons (approved):

PR #430, opened 2024-09-20
RAS-1296 Update Flask cors in secure-message

LJBabbage:
# What and why?
This PR update flask CORS due to a vulnerability

edit:  on update another issue has been found which there is no fix so has been ignored

# How to test?
CORS is only used in the app create, so the fact the test run says it all. You can however deploy and run the app and acceptance tests if you like (I did)
# Jira

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,13 +7,13 @@ start:
 	pipenv run python run.py
 
 lint:
-	pipenv check -i 70612 -i 70624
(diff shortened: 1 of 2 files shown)


anwilkie (approved):
Looks good to me.

matthew-robinson-ons (approved):
Tests work, app deploys and new ignore is appropriate

PR #434, opened 2024-11-05
Add codeql

mfcrocker:
# What and why?
This PR introduces codeql github code scanning to secure message v2. It also fixes a couple of vulnerabilities within the repo. 

# How to test?
Check that codeql has been introduced correctly

# Jira
RAS-1267


anwilkie (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
```
LJBabbage, line 1: This is old yaml and should be updated, checkout is now on version 4

secure_message/repository/retriever.py
```diff
@@ -340,8 +340,8 @@ def check_db_connection():
         try:
             SecureMessage().query.limit(1).all()
         except Exception as e:
-            logger.exception("No connection to database")
-            response = jsonify({"status": "unhealthy", "errors": str(e)})
+            logger.exception("No connection to database", errors=str(e))
```
LJBabbage, line 343: logging the error of a generic exception is not great

LJBabbage (changes requested):
Although needed the generic updating of a pipfile.lock, makes targeted testing difficult

secure_message/repository/retriever.py
mfcrocker: It's not but it's an improvement on returning it through the API (CodeQL hated that). It's also an improvement on just logging "No connection to database" which I feel doesn't help us a ton.

It might be necessary to have a more comprehensive look at logging in this service if we've got stuff like this in there :)

LJBabbage (approved):

PR #462, opened 2026-03-27
RAS-1839 Introduce scheduled job to mark closed conversations for del…

LJBabbage:
# What and why?
This PR introduces a cron job and an end point to mark conversation ready for deletion
# How to test?
Using the spinnaker pipline "Deploy Version and Helm Chart" deploy this PR with the config branch https://github.com/ONSdigital/ras-rm-config/pull/311, then test the cron job using different secure message states to make sure it works correctly (open, closed, old, new etc) 

Screenshot 2026-03-30 at 12 35 25

# Jira

### DIFF
--- a/secure_message/application.py
+++ b/secure_message/application.py
@@ -168,6 +168,7 @@ def _request_requires_authentication():
         request.endpoint is not None
         and "health" not in request.endpoint
         and request.endpoint != "info"
+        and request.endpoint != "threadmarkfordeletion"
         and request.method != "OPTIONS"
     )
 
--- a/secure_message/application.py
+++ b/secure_message/application.py
@@ -25,7 +25,12 @@ from secure_message.resources.messages import (
     MessageModifyById,
     MessageSend,
 )
-from secure_message.resources.threads import ThreadById, ThreadCounter, ThreadList
+from secure_message.resources.threads import (
+    ThreadById,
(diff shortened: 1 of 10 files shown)


secure_message/application.py
```diff
@@ -162,6 +168,7 @@ def _request_requires_authentication():
         request.endpoint is not None
         and "health" not in request.endpoint
         and request.endpoint != "info"
+        and request.endpoint != "threadmarkfordeletion"
```
LJBabbage, line 171: Agreed that it doesn't need a JWT

SteveScorfield (approved):
Fully tested and works as expected. Code changes look fine. Approved.

arroyoAle (changes requested):
Cron job marks the conversation for deletion correctly, but we should also have it remove the mark for deletion when the conversation is re-opened

LJBabbage:
> Cron job marks the conversation for deletion correctly, but we should also have it remove the mark for deletion when the conversation is re-opened

It's not a requirement on the card

arroyoAle (approved):
> > Cron job marks the conversation for deletion correctly, but we should also have it remove the mark for deletion when the conversation is re-opened
> 
> It's not a requirement on the card

Happy with this card as it is based on the card, we will need to circle back to do this

PR #465, opened 2026-04-21
RAS-1901 Stale Pipenv lock on Secure messaging

LJBabbage:
# What and why?
There was a subtle lock build change due to psycopg2-binary and a version of pipenv used by github, it has been fixed. I've had to completely regenerate the lock so there are other package updates, however only one, Marshmallow, caused issues. It now explicitly passes a data_key which was then causing our behave tests to fall over

# How to test?
Deploy this branch and test both manually and with acceptance tests

# Jira


warrenbaileyons (approved):
Tested with the acceptance tests and by logging into both frontstage/rops and sending messages back and forth.

SteveScorfield (approved):

PR #468, opened 2026-04-24
RAS-1878 Scheduled job to delete conversations marked for deletion

LJBabbage:
# What and why?
Creates an end point and endpoint to delete conversations (and related secure messages/statuses) that have been marke_for_deletion
# How to test?
Deploy this branch and helm chart (-p Deploy Version and Helm Chart ) then create a number of secure messages and responses, with at least one closed. For completeness I used both cron jobs sm-closed-conversations-mark-for-deletion and sm-closed-conversations-deletion. I changed the dates in the db of when conversations were closed and ran the former to change the mark_for_deletion status and the later to then delete. 

N.B There is still a gap which we need to discuss in a tech session. It's feasible a conversation could be re-opened in the 23 hours it has been marked for deletion and therefore should not be deleted. There is a number of ways we might want to tackle that

# Jira

### DIFF
--- a/_infra/helm/secure-message/values.yaml
+++ b/_infra/helm/secure-message/values.yaml
@@ -69,5 +69,9 @@ app:
 crons:
   closedConversationsMarkForDeletionScheduler:
     name: sm-closed-conversations-mark-for-deletion
-    cron: "0 0 * * *"
+    cron: "0 1 * * *"
     target: "/threads/closed/mark_for_deletion"
+  closedConversationsDeletionScheduler:
+    name: sm-closed-conversations-deletion
+    cron: "0 0 * * *"
+    target: "/threads/closed/delete"
--- a/secure_message/application.py
+++ b/secure_message/application.py
@@ -28,6 +28,7 @@ from secure_message.resources.messages import (
 from secure_message.resources.threads import (
     ThreadById,
     ThreadCounter,
+    ThreadDeletion,
     ThreadList,
     ThreadMarkForDeletion,
 )
@@ -68,6 +69,7 @@ def create_app(config=None):
     api.add_resource(ThreadById, "/threads/<thread_id>")
     api.add_resource(ThreadCounter, "/messages/count")
     api.add_resource(ThreadMarkForDeletion, "/threads/closed/mark_for_deletion")
+    api.add_resource(ThreadDeletion, "/threads/closed/delete")
 
     app.oauth_client_token_expires_at = maya.now()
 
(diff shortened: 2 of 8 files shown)


_infra/helm/secure-message/values.yaml
```diff
@@ -69,5 +69,9 @@ app:
 crons:
   closedConversationsMarkForDeletionScheduler:
     name: sm-closed-conversations-mark-for-deletion
-    cron: "0 0 * * *"
+    cron: "0 1 * * *"
```
LJBabbage, line 72: Marked for deleted at 01:00 and then deleted 23 hours later at 00:00

_infra/helm/secure-message/templates/cronjob_closed_conversations_mark_for_deletion.yaml
```diff
@@ -27,5 +27,5 @@ spec:
             args:
             - /bin/sh
             - -c
-            - curl -s -u $(SECURITY_USER_NAME):$(SECURITY_USER_PASSWORD) -X POST http://$(SECURE_MESSAGE_SERVICE_HOST):$(SECURE_MESSAGE_SERVICE_PORT)/$(TARGET)
+            - curl -s -u $(SECURITY_USER_NAME):$(SECURITY_USER_PASSWORD) -X POST http://$(SECURE_MESSAGE_SERVICE_HOST):$(SECURE_MESSAGE_SERVICE_PORT)$(TARGET)
```
LJBabbage, line 30: Noticed the double "/" as the value file already has it

secure_message/resources/threads.py
```diff
@@ -217,3 +217,16 @@ def post():
         except SQLAlchemyError:
             logger.exception("Failed to mark conversations for deletion")
             return "", 500
+
+
+class ThreadDeletion(Resource):
```
LJBabbage, line 222: At some stage this thread/conversation muddle needs correcting, but continued it for now

secure_message/application.py
```diff
@@ -169,6 +171,7 @@ def _request_requires_authentication():
         and "health" not in request.endpoint
         and request.endpoint != "info"
         and request.endpoint != "threadmarkfordeletion"
+        and request.endpoint != "threaddeletion"
```
LJBabbage, line 174: As agreed in tech, this doesn't need JWT

matthew-robinson-ons (approved):
Happy with this. Tested multiple rounds of conversation deletions in different data setups. Testing coverage looks strong and code changes are good.

arroyoAle (approved):
Looks good, tested with different combinations of secure messages and only the right ones get marked for deletion and deleted when the jobs run

PR #476, opened 2026-06-15
RAS-1978 Optimise ROPS query

LJBabbage:
# What and why?
The status table is over processing on FDI, it pulls in 4 different surveys and then overloops due to the different states it can be in inbox, sent and unread. This reduces the looping in status table. 

# How to test?
Make sure SM still works
# Jira

https://officefornationalstatistics.atlassian.net/browse/RAS-1978

### DIFF
--- a/secure_message/repository/retriever.py
+++ b/secure_message/repository/retriever.py
@@ -408,6 +408,7 @@ class Retriever:
         where_clauses = list(filters["secure_message_filters"])
 
         if is_actor_query:
+            # As we don't use any data from the status table, we only need to check a record exists for the actor.
             where_clauses.append(f"""
             EXISTS (SELECT 1 FROM securemessage.status s
                     WHERE s.msg_id = sm.msg_id AND {filters["actor_filter"]})
--- a/secure_message/repository/retriever.py
+++ b/secure_message/repository/retriever.py
@@ -401,17 +401,6 @@ class Retriever:
                 sm.sent_at
             """
 
-        if is_actor_query:
-            from_sql = """
-            FROM securemessage.status s
-            INNER JOIN securemessage.secure_message sm
-                ON sm.msg_id = s.msg_id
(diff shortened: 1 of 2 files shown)


secure_message/repository/retriever.py
```diff
@@ -401,31 +401,23 @@ def _build_thread_query(filters: dict, is_count_query: bool = False, is_actor_qu
                 sm.sent_at
             """
 
-        if is_actor_query:
```
LJBabbage, line 404: Always link to the secure message table first

warrenbaileyons:
I've run the acceptance tests and they've passed. I've also logged into both frontstage and response operations and done the following:
1. Send a message internal to external
2. Send a message external to internal
3. Close the conversation
4. View closed conversation internal
5. View closed conversation external
6. Reopen conversation

secure_message/repository/retriever.py
```diff
@@ -401,31 +401,23 @@ def _build_thread_query(filters: dict, is_count_query: bool = False, is_actor_qu
...
         conversation_filters = (
             " AND ".join(filters["conversation_filters"]) if filters["conversation_filters"] else "TRUE"
         )
         latest_sm_sql = LATEST_SM_SQL.format(conversation_filters=conversation_filters)
         where_clauses = list(filters["secure_message_filters"])
 
         if is_actor_query:
-            where_clauses.append(filters["actor_filter"])
+            where_clauses.append(f"""
+            EXISTS (SELECT 1 FROM securemessage.status s
```
warrenbaileyons, line 413: It might be worth a comment to explain we just need to check that there is a status entry in the table for the user that is looking at this conversation

warrenbaileyons (approved):
LGTM, discussed with Lewis and I'm happy with it

SteveScorfield (approved):

PR #482, opened 2026-08-06
Rename DATABASE_URL to DATABASE_URI and share postgres image with CI 

prem-buczkowski-ons:
# What and why?
DATABASE_URI is the convention across the other python services (6 of 7); this
was the only outlier. The chart assembles the value from the DB_* secrets rather
than reading a secret of this name, so the rename is self-contained.

The postgres version now lives in _infra/postgres-image, fed to docker-compose
via --env-file so local dev and CI share it, and bumped from the EOL 9.6 to 14.
Compose requires that flag, so use `make start-db` rather than `docker compose up`.

All of this is so the common Cloud Build pipeline we are developing for all Python
services has as few exceptions needed per-service as possible.

# How to test?
Run `make start-db start` and check if the service is working.

# Jira
https://officefornationalstatistics.atlassian.net/browse/RAS-1930
https://officefornationalstatistics.atlassian.net/wiki/spaces/SDC/pages/303956413/Spinnaker+Removal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,11 @@
-.PHONY: build start
+.PHONY: build start start-db
 
 build:
 	pipenv install --dev
+
+# The postgres image version is read from _infra/postgres-image, which CI uses too.
+start-db:
+	docker compose --env-file _infra/postgres-image up -d
 
 start:
 	pipenv run python run.py
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,20 +24,6 @@ jobs:
   build:
     name: Build & Package
     runs-on: ubuntu-latest
-    services:
-      # Label used to access the service container
-      postgres:
-        # Docker Hub image
-        image: postgres:9.4
-        env:
-          POSTGRES_PASSWORD: postgres
-          POSTGRES_USER: postgres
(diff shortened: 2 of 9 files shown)


warrenbaileyons:
Tested on my dev environment and URI change works

Makefile
```diff
@@ -1,8 +1,12 @@
-.PHONY: build start
+.PHONY: build start start-db
 
 build:
 	pipenv install --dev
 
+# The postgres image version is read from _infra/postgres-image, which CI uses too.
+start-db:
+	docker compose --env-file _infra/postgres-image up -d
+
```
warrenbaileyons, line 11: same as the other PR, we'll need a target for the Macbooks that use podman

warrenbaileyons: It might be worth us add a start-local-db target that uses `podman compose  --env-file _infra/postgres-image up -d`

warrenbaileyons:
LGTM, lets add that extra target and then we can approve/merge 👍

Makefile
prem-buczkowski-ons: Done.

warrenbaileyons (approved):

arroyoAle (approved):

